PR #94, opened 2014-06-03
Support collection level group permissions

russellhutsonmicro-p:
Added simple support for controlling access to collections.
It creates the necessary resource for each collection. This then appear on the Settings->Accounts->Groups->Access tab and can be selected or deselected within each Group.


Tails:
Does this include access levels per entry locale?

PR #214, opened 2014-11-15
Mistake in function name updateSope.

thinkink:
Mistake in function name updateSope. changed to updateScope.


aheinze:
already fixed

PR #223, opened 2014-12-10
Mailer: Adding fromName to config and easier debug for mail options.

piotr-cz:
There are 2 features in this this PR:
- first commit allows adding a `fromName` optional option to mailer config in `cockpit/custom/config.php` file. I'll update [docs](https://github.com/aheinze/cockpit-docs/blob/master/getting_started/configuration.md) when it's merged.
```php
return [
    'mailer' => [
        'from'     => 'santa@gmail.com',
        'fromName' => 'Santa Claus', // New option
    ],
];
```
- the second one allows easier debug for mailer options. It tries to submit testing email allowing PHPMailer to throw exceptions, catches it and then shows up to user. It's much easier to see now what's wrong (host unavailable, user unavailable, password is wrong): ` Sending email failed: (exception message)`

Sorry for submitting these two in one PR.
I've got more improvements coming to mail and squashed these by mistake.


piotr-cz:
So I'm in the same place now, trying to debug my mail configuration but all I get is just a 
_Sending email failed._ message.
Please consider this pull request.

aheinze:
thanks!

PR #231, opened 2014-12-15
Allow adding attributes to form element and disabling ajax validation

piotr-cz:
Features:
- Option to disable default ajax form validation by passing `"ajaxValidation" => false` parameter.
- Option to pass any form element attribute (`role="form"`, `ng-controller="ContactCtrl as ctrl"`) by 3rd parameter. _I left the class attribute in 2nd group for backward compatiblity_.

``` php
<?php cockpit(
   "forms:form",
   "contact",
    ["ajaxValidation" => false],
    ["role" => "form", "ng-controller" => "ContactCtrl as ctrl"]
) ?>
```

### DIFF
--- a/modules/core/Forms/bootstrap.php
+++ b/modules/core/Forms/bootstrap.php
@@ -102,15 +102,16 @@ $this->module("forms")->extend([
         return $forms[$name];
     },
 
-    "form" => function($name, $options = []) use($app) {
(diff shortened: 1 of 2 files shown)


aheinze:
thanks for the contribution...custom attributes is on my todo list. I'll take your implementation as a reference.

PR #247, opened 2015-01-07
DirectoryIndex index.php

1008k:


timaschew:
I realized that after a fresh installation of the latest stable version of Cockpit, on a Apache with PHP 7.0 I got all the time a forbidden page. It worked if I manually add the `index.php`.
After reinstallation with PHP 5.6 it worked. Maybe this will resolve the issue with PHP 7

PR #296, opened 2015-04-04
Fixed wrong created date on datastore items

florianletsch:
Was wondering why the created date on all my datastore items was the same. This fixes it.


aheinze:
thx!

PR #298, opened 2015-04-15
Fix update of the image title field for a gallery item

jorisvandesande:
Because the title was not saved after clicking the save button.


aheinze:
thanks!

PR #307, opened 2015-05-05
Same breadcrumb in Datastore config as in other modules

piotr-cz:
Same breadcrumbs markup as on other pages (ie. [Accounts page](https://github.com/aheinze/cockpit/blob/0.13.0/modules/core/Auth/views/accounts/index.php#L5)).

I wasn't sure about this one, as some other pages ([General](https://github.com/aheinze/cockpit/blob/0.13.0/modules/core/Cockpit/views/settings/general.php#L8), Addons) use h1


aheinze:
thanks!

PR #310, opened 2015-05-10
display addon info

malles:
Display the addon info in the table.
Maybe pull this in, or just use as inspiration for the big rewrite.

Thanks!


aheinze:
thanks!

PR #324, opened 2015-06-08
Flush output buffering for GZIP response

piotr-cz:
There's an undocumented feature in cockpit for compressing the output,
however it's missing the function to flush output buffering, so browser ends up with invalid response.

In my opinion this functionality could be dropped as it's easier to setup compression in `.htaccess` file which is required anyway by default.

### DIFF
--- a/vendor/Lime/App.php
+++ b/vendor/Lime/App.php
@@ -878,7 +878,11 @@ class App implements \ArrayAccess {
 
             $self->trigger("after");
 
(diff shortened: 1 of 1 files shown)


lfbittencourt:
The only way I've fixed this annoying error was disabling deflate module. Why don't you merge it?

piotr-cz:
This is a question to repo maintainer, @aheinze

PR #334, opened 2015-07-12
Fix: WYSIWYG editor is inserting relative URLs

charkes:
Fixes aheinze/cockpit#217. TinyMCE was overwriting Cockpit’s media URLs.

http://www.tinymce.com/wiki.php/Configuration:relative_urls


charkes:
I added 2 more commits to fix small bugs.

aheinze:
thanks!

PR #335, opened 2015-07-13
Add Countable Interface to MongoLite Cursor

piotr-cz:
This helps with checking whether collection result set is empty, by `count($collection)`.

_I use Twig for rendering results set and it's `empty` expression checks if given object implements Countable interface. If so, returns count($collection);_

Since the Cursor class has already a [count](https://github.com/aheinze/cockpit/blob/0.13.0/vendor/MongoLite/Cursor.php#L67) method, all that needs to be done is to implement [Countable](http://php.net/countable) Interface

### DIFF
--- a/vendor/MongoLite/Cursor.php
+++ b/vendor/MongoLite/Cursor.php
@@ -5,7 +5,7 @@ namespace MongoLite;
 /**
  * Cursor object.
  */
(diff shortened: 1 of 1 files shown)


aheinze:
thanks!

PR #336, opened 2015-07-14
Can't make Collection boolean field default true

charkes:
I was unable to make a boolean default to true as it would store a 1 integer or a string of “true”.


aheinze:
thanks!

PR #338, opened 2015-07-15
Option for slugs to be unique

piotr-cz:
I'm using entry slugs in routes.

After this pull request when field type is set to _slug_, shows up another option to set it to _unique_.

Then when user tries to save an entry with a slug that is already present in collection, API returns a message `There is already an entry in this collection with this slug for field '[slugname]'.` and entry is not saved.


aheinze:
thanks!

PR #346, opened 2015-07-25
Auto-orient uploaded photos

charkes:
Photos from (for example) iPhones will be rotated all different directions. This is corrected on upload so all previews are rotated to the correct orientation. Subsequent thumbnail/crop actions will use the correct orientation.

PHP EXIF must be enabled for the code to read the orientation.

Fixes aheinze/cockpit#339


aheinze:
wouldn't it be better to check it it is an image first before using the exif function?

charkes:
Good idea.  I'll implement that and add another commit.

aheinze:
:+1:

Issue #343, opened 2015-07-20
Get the prev/next posts in JS


manudurgoni:
Hey,

I really appreciate Cockpit !

But I've a problem to get the prev/next posts from a current entry.

I'd like to have the posts just before and after my entry with _id `_this.actualite._id`

I've found in the mongoDB docs the `$lt`and `$gt` but how I can use those properties ?

I've tried this, but it doesn't work :

``` js
Cockpit.request('/cockpit/call', {
    module: 'collections',
    method: 'find',
    args: [
       'Actualites', {
          filter: {
             'categorie': _this.categorie._id,
              _id: {
                 $lt: _this.actualite._id
              }
           },
           limit: 1,
           sort: {
             _id: -1
           }
        }]
        }).success(function(item) {
          console.log(item);
          console.log(item[0].titre);
      });
```

Maybe someone can help me :)

Thanks

piotr-cz:
Why won't you load whole collection (respecting the order), lookup current entry and pick previous/ next entries from an array?
That's how it's done for example in [Joomla CMS](https://github.com/joomla/joomla-cms/blob/3.4.3/plugins/content/pagenavigation/pagenavigation.php#L172). You could also add a _collection-link_ type field and set up these manually.

The `$lt` and `$gt` are comparators when picking up entries based on criteria, I think it's better to use them in conjunction with an custom-order field.
BTW: I have bad experience with filtering the `_id` field, Cockpit handles this propety in a different way.

manudurgoni:
Yep, I thought to your solution, but as you wrote, finally I needed to find previous / next entries based on a custom field (a date selected by editor).


```js
// Previous entry
Cockpit.request('/collections/get/Entries', {
  filter: {
    'date': {
      '$lt': _this.entry.date
    },
    'category': _this.category._id
  },
  limit: 1,
  sort: {
    'date': -1
  }
})

// Next entry
Cockpit.request('/collections/get/Entries', {
  filter: {
    'date': {
      '$gt': _this.entry.date
    },
    'category': _this.category._id
  },
  limit: 1,
  sort: {
    'date': 0
  }
})

```

At first it did not work, but I noticed that the $lt / $gt filters in MongoLite did not allow to compare a string (and my entry's date was a string).

So I fixed this in MongoLite to solve my problem:


```php
// cockpit/vendors/MongoLite/Database.php
private static function evaluate($func, $a, $b) {
  $r = false;

  switch ($func) {

    ...

    case '$gte':
[...]

PR #347, opened 2015-07-28
Removing is_numeric checks in gte, gt, lte, lt comparisions

piotr-cz:
Because in PHP you can compare anything, see PHP.net: [Comparision operators](http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.types).

This is helpful when comparing dates stored in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format: `(true) '2015-06-06' > '2015-01-01'`.

Also this is more in line with the `$eq` and `$not` operators which ATM don't check if compared values are numeric.

Reference: Issue #343

### DIFF
--- a/vendor/MongoLite/Database.php
+++ b/vendor/MongoLite/Database.php
@@ -256,32 +256,27 @@ class UtilArrayQuery {
             case '$eq' :
                 $r = $a == $b;
                 break;
+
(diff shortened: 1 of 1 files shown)


aheinze:
thanks!

PR #349, opened 2015-08-04
Added NOT operators to MongoLite's $in and $has

charkes:
I added $nin for 'not in' and $nhas for 'not has'.


aheinze:
Thanks! Could you also please create a pull request for the main MongoLite repo here: https://github.com/aheinze/MongoLite

PR #376, opened 2015-12-25
Display image filename while editing meta in gallery

attitude:
Help user when editing gallery image caption, width, height or other meta by displaying the filename in modal header.

Result:
screen shot 2015-12-25 at 19 01 41


aheinze:
Thanks!

PR #378, opened 2015-12-30
Fix `undefined` ngModel for tag directive in gallery view

attitude:
**Issue:** Loading image tag data into modal after click fails because modal and fields are compiled before any ngModel is assigned to the tag field directive.

Solution: Create scope using `ng-if` on modal and compile modal fields after `ng-if`evaluates as `true` and creates new scope. That also requires changing modal behaviour on open (modal DOM is created after the click) and on close (binding close buttons, esc key to scope method).

*Note: there's 300ms timeout to finish modal hide transition according to CSS.*

See the XHR preview of `data.images[0].data.technique`.

Before:
Before

After:
After


aheinze:
thanks!

PR #388, opened 2016-02-27
Location field tweaks

piotr-cz:
- Keep location address when using geolocation and save it in entry, so it's possible to reuse it in later.
  When user changes location using mouse, input value is reset to `''`.
- Use current Cockpit locale for Google Maps

cockpit-location

This is for Cocpit standard. If all is good I'll prepare commit for Cockpit-Next.


aheinze:
Thanks!

PR #389, opened 2016-02-29
made collections queryable by slug

gryphonmyers:


piotr-cz:
Can you describe how to use it?

gryphonmyers:
Only the first commit was supposed to be associated with this PR, my bad. Second feature is not quite done yet. I should have made a different branch.

Queryable by slug feature is super simple - I just made it so rest API will accept both

```
/cockpit/rest/api/collections/get/MyCollectionName?token=xoxoxo
/cockpit/rest/api/collections/get/mycollectionname?token=xoxoxo
```

the benefits of this become evident when the name contains whitespace:

```
/cockpit/rest/api/collections/get/My%20Collection%20Name?token=xoxoxo //<- I couldn't even get this working
/cockpit/rest/api/collections/get/my-collection-name?token=xoxoxo //(yay! now we can access collections with whitespacey names)
```

The second feature is a flag that can be added to any rest api to denormalize linked collections. For instance, before a rest api call might look like:
`
/cockpit/rest/api/collections/get/MyCollection?token=xoxoxo`

`
[{
  "Name": "My Entry",
  "MyLinkedCollection": "xxxxxxxxxxxxxxxxxxxxxxxx"
}]
`

Now, we can do this:

`
/cockpit/rest/api/collections/get/MyCollection?token=xoxoxo&denormalize=true`

`
[{
  "Name": "My Entry",
  "MyLinkedCollection": {
    "Name": "MyLinkedEntry"
    "_id": "xxxxxxxxxxxxxxxxxxxxxxxx"
  }
}]
`

This feature is almost done but missing some hashes in the recursive process so I will be debugging that today.

piotr-cz:
great, thanks.
Yes, you should probably create new PR for other feature

aheinze:
To be honest,

I would prefer a more explicit approach:

```
public function getBySlug($slug=null) {
    //...
}
```

piotr-cz:
Me too, because there might be two collections: one with the same name as the slug of another.

gryphonmyers:
Good point - I'll work on this

gryphonmyers:
How's that?

aheinze:
@gryphonmyers sorry for the late answer. but I think we're still not there. Your solution can cause some side effects if we call the get method with a different parameter the **name** or **slug**. I would suggest:

```php
public function getBySlug($collection=null) {
    return $this->get($collection, true);
}

public function get($collection=null, $bySlug = false) {

    if (!$collection) {
        return false;
    }

    $findBy = $bySlug ? 'slug':'name';
    
    $collection = $this->app->db->findOne("common/collections",  [$findBy=>$collection]);
    if (!$collection) {
        return false;
    }
    $entries = [];
    if ($collection) {
        $col     = "collection".$collection["_id"];
        $options = [];
        if ($filter = $this->param("filter", null)) $options["filter"] = $filter;
        if ($limit  = $this->param("limit", null))  $options["limit"] = $limit;
        if ($sort   = $this->param("sort", null))   $options["sort"] = $sort;
        if ($skip   = $this->param("skip", null))   $options["skip"] = $skip;
        $entries = $this->app->db->find("collections/{$col}", $options);
    }
    return json_encode($entries->toArray());
}
```

gryphonmyers:
Ok that works

aheinze:
thanks

Issue #322, opened 2015-06-02
Media Path not saving in groups


gazsp:
I'm using the master branch and MongoDB, and the following gets saved each time to the `cockpit_settings` collection:

``` json
{
    "_id" : "...",
    "key" : "cockpit.acl.groups.settings",
    "val" : {
        "admin" : [],
        "users" : []
    }
}
```

I think it's a problem on the Javascript side, as `admin` and `users` should be objects, not arrays (I assume).

toish:
I'm having this problem also, an empty array is being sent to saveAci, so it's a JS thing.
<img width="964" alt="screen shot 2016-02-09 at 5 47 53 pm" src="https://cloud.githubusercontent.com/assets/8439945/12933305/672ba250-cf55-11e5-8787-5b7d97797e5d.png">

PR #390, opened 2016-03-01
Fixed issue #322, root media path can now be saved

Michaelvilleneuve:
I just specify the default root value instead of passing an empty object which blocked saving.
Please note that there is probably a better way, but still, it doesn't even add a line. :)


aheinze:
Thanks!

PR #391, opened 2016-03-02
Location field tweaks for Cockpit-Next branch

piotr-cz:
see #388 

Additionally removed the `sensor=false&` param from google maps api loader, to prevent warning _SensorNotRequired_ (see https://developers.google.com/maps/documentation/javascript/error-messages#sensor-not-required)


aheinze:
Thanks!

PR #392, opened 2016-03-02
Build scripts

piotr-cz:
Add nodejs modules to package.json, so it's possible to run build scripts without having them installed globally.

Also added information in README.md file:

### Build

You need [nodejs](https://nodejs.org/) installed on your system.

First run `npm install` to install development dependencies

a) Run `npm run build` - For one-time build of styles and components
b) Run `npm run watch` - For continuous build every time styles or components change

___

ps: What about adding frontend assets (riot and uikit) too?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /config/*
 /modules/addons/*/*
 /tests/*
+/node_modules
 
 !.gitignore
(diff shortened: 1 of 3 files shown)


aheinze:
Ah yes, this a great addition! Thanks!

PR #394, opened 2016-03-08
Choose which Collection-link field to show instead of collection Id

piotr-cz:
When using Linked-collection, this allows to choose which field should be shown in entries page instead of an Id.

You may choose any required field which is able to have string representation (is not type of location, gallery, multifield, link-collection, media or region).

Multiple collections are supported.

### Example

Let's say we have two collections: Countries and Nodes with relationship 1:n

When configuring Nodes collection we create an Instance field, set type to Collection link and set new option _Field on entries list page: Name_

Before:
before

After:
after


aheinze:
Thanks!

piotr-cz:
When thinking about the implementation, there will be a problem when one collection has two fields of type Collection-link which are linking into the same target collection.

The later one will overwrite data of the first one.

Linked collections data are cached in a double-nested hashmap (see schema below) and when main collection entries are loaded, these data are populated.

```JSON
{
    "[target-collection-id]": {
        "[target-entry-id]": "[target-entry-selected-field-value]"
    }
}
```

I think it's safe to consider it an edge case for the moment.

One solution is to to change format of retrieved data, other is to populate data already at the API level similarly to how it's done in the [module API](https://github.com/aheinze/cockpit/issues/54#issuecomment-48674286)

I'm attaching a screenshot of extended Collection-Link field used in description example

extended-collection-link-field

PR #397, opened 2016-03-17
fixed error involving php notice and toString on field

### LINKED ISSUE #396: Entries page doesn't list entries
@piotr-cz 

So I attempted to test 700b75a but it appears to have the effect of emptying the entries page:

Without new code: 
image

With new code:
image

I don't understand the context of this code well enough to spot the error. It appears to occur when any field within the collection has an entries field other than Collection ID set.

gryphonmyers:
Fixes #396

### DIFF
--- a/modules/core/Collections/Controller/Api.php
+++ b/modules/core/Collections/Controller/Api.php
@@ -114,7 +114,7 @@ class Api extends \Cockpit\Controller {
             $col     = "collection".$collection["_id"];
             $options = [];
 
-            if ($collection["sortfield"] && $collection["sortorder"]) {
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks!

piotr-cz:
Thanks. Till now I've always had to include the `sortfield` parameter, so I've been using such workaround:

```JS
App.route('/api/collections/entries'),
{
    collection: {
        _id      : collection._id,
        sortfield: null
    }
},
```

piotr-cz:
Instead of removing the `.toString()` it's better to not process field that contains null values as collection id's at all - see #398.

But it's fine to do both things to be safe. Values are stringified later anyway inside the `.join(', ')` concatenator which doesn't throw errors when joining array of undefined's

PR #398, opened 2016-03-18
Fix showing Linked collections values when value is not showable

piotr-cz:
Handling scenario when collection ID is not filled in.

Fixes frontend part of #396


aheinze:
Thanks!

PR #399, opened 2016-03-20
Fix app-header z-index

piotr-cz:
Fixed menu z-index.

Attaching screenshot how it looks for me on Chrome v49.0.2623.87 m (64-bit)

cockpit-zindex

In the fix I gave `.app-header` z-index of 10, but maybe a lower z-index should be also set on `.app-main` and `.app-footer`


piotr-cz:
Fixed in a7eb6187a13d1b1810340deff8b18963f71fe808

PR #400, opened 2016-03-21
Collection links improvements

piotr-cz:
- Show field name if label is not filled in. At the moment it shows an empty <option> element
- Allow picking field which are not marked as required.
  When their value is not set, it's just not shown. It was pointless restriction and didn't really work
- Fix multiple collection-links (partial revert of https://github.com/aheinze/cockpit/pull/398/commits/b55400cd57e296fa623b682fec51f5f6f862439b)

This is extension to #394

### DIFF
--- a/modules/core/Collections/assets/js/collection.js
+++ b/modules/core/Collections/assets/js/collection.js
@@ -107,9 +107,9 @@
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


aheinze:
Thanks!

PR #403, opened 2016-03-27
Display locations address on entries page

### LINKED ISSUE #402: Bug: Location Field in Listed Entries Display as "[object Object]"
So, I've added in a location field in and filled out an entry with a location. However, once I saved it, the location field in the entry doesn't list the location or even the coordinates. It only says `[object Object]`.
screen shot 2016-03-25 at 16 04 08

piotr-cz:
Fixes #402 
Better solution would be to implement a `toString` method inside location field type (and all other field types) and apply it on fetched entry data, but that is something to discuss.

I've got same fix for master branch too but I'll see if there are any comments for this one.

### DIFF
--- a/modules/core/Collections/views/entries.php
+++ b/modules/core/Collections/views/entries.php
@@ -208,6 +208,18 @@
 
             return App.callmodule('collections:find', [this.collection.name, options]).then(function(data){
 
+                data.result.forEach(function(entry) {
+
+                    // Process fields
+                    for (var key in entry) {
+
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks! But I took a more general approach for now: https://github.com/aheinze/cockpit/commit/1a9d10f9794aaceeefd41f1cdef4ddcb0341c351

piotr-cz:
That's definitely more bullet-proof.

I still think it makes sense to format entry data into user readable form, I'll elaborate on universal approach.

PR #404, opened 2016-03-29
triggered js events for collection save and remove

gryphonmyers:
Added a couple event hooks (I am writing an add-on that depends on these).


aheinze:
Thanks!

PR #413, opened 2016-04-30
added a check on field names that cannot have spaces: user is alerted

zontarian:
if users create a field in a Collection with a malformed name, nothing happens, the field is created and can be edited, but there is an error during data binding because the field name is used verbatim as a JS attribute to some object. 

If the field name has, for example, spaces, the data that the user inputs is never bound to underlying object and the field is submitted empty to the Save Rest API. No error is risen so the user is not aware that something went astray.

The simplest workaround was to validate each keystroke while editing the Field name, and alert the user that spaces (whitespaces) cannot be inserted. Spaces are not inserted and a visual "flash" appears in the field, as well as an alertify (just once).


zontarian:
Stupid me, maybe a `pattern="[a-zA-Z]?[a-zA-Z0-9_]"` should be easier? :)

aheinze:
I would prefer the pattern solution ;-)

PR #414, opened 2016-04-30
added feature to pass along codemirror options to markdown editor via…

zontarian:
I'd like some more flexibility on the markdown editors, and since there is an "options" field when creating a new field for a collection, I thought to use it to pass some options to UIKit and all throught Codemirror.

I'm not sure this is the best way yo do it, but it works. I'm not very proud because I had to put an extra tag in the `field-markdown.tag` but it works.

At the moment I pass all the properties in the `codemirror` block to codemirror via  the `opts` object in this call to `UIkit.htmleditor(this.input, opts)`

To test it, create a `markdown` field in a collection, and in the options put this JSON

`{
  "codemirror": {
    "mode": "split",
    "height": 250,
    "maxsplitsize": 800
  }
}`

The resulting field, while editing or adding a new entry for the collection, will be a shorter field, with a split panel.

### DIFF
--- a/modules/core/Cockpit/assets/components/field-html.tag
+++ b/modules/core/Cockpit/assets/components/field-html.tag
@@ -37,7 +37,14 @@
             ], function() {
 
                 $this.input.value = $this.value;
-
+                //pass along codemirror options to htmleditor/codemirror
+                if(opts.codemirroroptions){
+                    co=opts.codemirroroptions;
+                    for(prop in co){
+                        if(!(prop in opts))
+                            opts[prop]=co[prop];
+                    }
+                }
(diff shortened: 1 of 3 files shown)


modules/core/Cockpit/assets/components/field-html.tag
```diff
@@ -37,7 +37,14 @@
             ], function() {
 
                 $this.input.value = $this.value;
-
+                //pass along codemirror options to htmleditor/codemirror
+                if(opts.codemirroroptions){
+                    co=opts.codemirroroptions;
```
piotr-cz: This line should be `var co = opts.codemirroroptions;`, otherwise it's leaking into global scope

PR #423, opened 2016-07-20
Option to set google masp API key (Cockpit-Master)

piotr-cz:
As described in [this announcement](http://googlegeodevelopers.blogspot.com.au/2016/06/building-for-scale-updates-to-google.html), Google Maps requires using API key for domains active after 22/06/2016 and localhost domain.
Without it, widget doesn't render and console shows a `MissingKeyMapError`.

I've added new tab: _Settings > General > Google Maps_ to have add an ability to specify the API key.
Data is stored under collection `cockpit/settings` and key `cockpit.gmaps.key`.

gmapskey2

This PR concerns (old/ master) Cockpit branch, possibly similar implementation may be prepared for Cockpit Next (effort started [here](https://github.com/COCOPi/cockpit/commit/368a4d8d4f5bf277187e6b625bcfdc9d5fb021c5))

### DIFF
--- a/assets/js/angular/fields/location.js
+++ b/assets/js/angular/fields/location.js
@@ -4,8 +4,7 @@
 
 (function($){
 
-    angular.module('cockpit.fields').directive("locationfield", ['$timeout', '$compile', function($timeout, $compile){
-
+    angular.module('cockpit.fields').directive("locationfield", ['$timeout', '$compile', '$http', function($timeout, $compile, $http) {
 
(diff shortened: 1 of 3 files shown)


aheinze:
Thanks!

piotr-cz:
I was wondering this should go to the Settings/ General or Api Access.
I guess in the Cockpit Next Google maps API key could be set in System settings file

aheinze:
right ;-)

PR #429, opened 2016-08-12
Fix bug in rest-api.php

joejordanbrown:
Correct Mime type changed to JSON.


aheinze:
Thanks for the contribution, I would like to suggest to check the output type before setting the mime type.

joejordanbrown:
Actually good idea. Ok do you have a built in function in lime for checking data types? I don't mind doing another pull.

PR #434, opened 2016-08-31
Add controls for geolocation adjustments (master branch)

piotr-cz:
Sometimes Google geolocation mechanism doesn't find correct coordinates of an address.
This PR adds controls to tweak it; it adds controls to:

- Specify points' latitude and longitude using manual inputs
- Options to:
  - Keep the text address while changing point location
  - Reverse geocode (find address by points' location)

location-manual


aheinze:
thanks!

PR #436, opened 2016-09-08
Tweaks to importer

gryphonmyers:
I had identified two problems with the importer:

1) It only accepted arrays, while the legacy version of Cockpit outputs objects.
2) It did not create a complete list of column headers from the input object. Previously we were only checking the first item in the collection, however the legacy version of Cockpit sometimes outputs items that do not contain keys for optional fields; some items may have keys that others do not.

These commits should address both of these issues.

Additional changes:

Under some circumstances, the importer was incorrectly establishing collection links with the first entry in the collection. That is fixed.

I have also added the ability for the collection link import filter to grab lookup value out of an object, if the same key exists there.

### DIFF
--- a/modules/Collections/assets/import/filter.js
+++ b/modules/Collections/assets/import/filter.js
@@ -3,7 +3,6 @@
     var Filter = {
 
         filter: function(field, value, extra){
-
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks!

PR #444, opened 2016-09-12
[Next] added date and time import filters

gryphonmyers:
Added import filters for coercing unix timestamps to Cockpit's separate date and time formats.


gryphonmyers:
Replace with #447

PR #446, opened 2016-09-13
[Next] Add link property to collection link object

gryphonmyers:
I was finding it impossible to denormalize the collection links from data in the REST API, so I have passed through the value from "link" option, so a secondary call may be made to denormalize.

### DIFF
--- a/modules/Collections/assets/field-collectionlink.tag
+++ b/modules/Collections/assets/field-collectionlink.tag
@@ -196,10 +196,10 @@
 
         var _entry = e.item.entry;
         var entry = {
-+           _id:_entry._id,
-+           display: _entry[opts.display] || _entry[this.collection.fields[0].name] || 'n/a',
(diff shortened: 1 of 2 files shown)


aheinze:
but you have the link info already in the collection definition, this looks redundant to me

gryphonmyers:
Yes but if collection A contains a collection link to collection B, when fetching collection A from the REST API the only information that comes through for that collection link is:
`
{
  _id:"xxxxx",
  display: "Entry Name"
}
`
The link info exists in the collection definition but I don't see that from the REST API. Without this information, I cannot form a request for the linked entry.

Instead of storing this in the link, it could be inserted into the REST response(?) That seemed like a more involved change though

aheinze:
I updated the collection api request result. the response will now look like this:

```json
{
   "fields": {...},
   "total": <number>,
   "entries": [...]
}
```

PR #450, opened 2016-09-14
[Next] added bare entries flag to rest api

gryphonmyers:
For those who would prefer the old response format


aheinze:
Is this just a preference or is there really a need for this option?

gryphonmyers:
The main concern is reverse compatibility, but I suppose as an opt-in flag that's out the window anyway. Personally, yes I would find it useful so I don't have to adjust the way I am using the API in some implementations, but instead just add one url param.

aheinze:
https://github.com/COCOPi/cockpit/blob/next/modules/Collections/Controller/RestApi.php#L22-L25

PR #472, opened 2016-10-05
added support for associative languages array in confg

gryphonmyers:
This allows for language config like:
```
"languages" => [
        "German"=>"de",
        "Spanish"=>"es"
]
```
The key will show up in the select while the value continues to be used internally.

### DIFF
--- a/modules/Collections/views/entry.php
+++ b/modules/Collections/views/entry.php
@@ -92,11 +92,11 @@
                     <div class="uk-width-1-1 uk-form-select">
 
                         <label class="uk-text-small">@lang('Language')</label>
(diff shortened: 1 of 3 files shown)


aheinze:
Cockpit now supports:

```
"languages" => ["de", "es"]
```

or

```
"languages" => [
        "de" => "German"
        "es" => "Spanish"
]
```

piotr-cz:
How do I use this feature in collection entries?
I can switch languages but only thing that changes is the revision count.
I've tested with default _Posts_ template and when peeking into the _collections.sqlite_ only the the default language data are stored

piotr-cz:
Fix in #547

PR #482, opened 2016-11-13
Fixed Collections.filterNonLinkFields which was preventing populate from working.

synthecypher:
I think this must have just been a typo, its working fine after this change.


aheinze:
Thanks!

PR #499, opened 2016-12-13
Ability to query all collections at once & some minor fixes in webhooks

mathielen:
Hi there,

I made some enhancements that I needed.

- There was an issue with the webhook url had an empty character in front. Fixed it with trimming the url
- If the webhook was a SSL-URL the curl call failed, fixed it
- If the webhook fails for some reason - trigger an error
- Added the ability to query all collections at once with the API. So Instead of calling /api/collections/get/<collection>?field=value you now can also do /api/collections/get?field=value and the api returns all entries that matches field=value in any collection

Best,
Markus


aheinze:
Thanks for the contribution! I'll merge the webhook fixes, but need to think about the query all collections code.

aheinze:
I close this one. to query multiple collections I would recommend using the following addon: https://github.com/agentejo/CockpitQL

PR #516, opened 2017-02-25
Updated Lime to the latest version

ostgals:
Also Auth module was fixed to be compatible with the latest version of
Lime. Required version of PHP should be elevated to >=5.5


aheinze:
Thanks for the contribution, but as this is a PR for the legacy master branch it may break existing installations. So I don't feel comfortable to merge it :-/

ostgals:
That's OK Artur. Maybe it's time to completely divide master and next branches into separate repos. I very appreciate your efforts. Thanks and have my unpretentious donation!

PR #523, opened 2017-03-27
Add hasaccess check in forms dashboard widget

piotr-cz:
ATM forms widget shows an action _Create a collection_ even if user doesn't have rights to create one.

PR adds similar check like how it's done in other modules widgets (for example [in collections](https://github.com/COCOPi/cockpit/blob/next/modules/Collections/views/widgets/dashboard.php#L60-L64)).


aheinze:
thanks!

PR #528, opened 2017-04-11
Form entries display name and value the other way around

piotr-cz:
Before fix
form-before

After fix
form-after


aheinze:
Thanks!

PR #532, opened 2017-04-13
replaced instances of (apparently broken) bind attribute with native …

gryphonmyers:
…event attributes

Addresses https://github.com/COCOPi/cockpit/issues/530


aheinze:
Thanks!

PR #535, opened 2017-04-17
fix riot dependency

timaschew:
otherwise it's depending on the implementation of resolving and installing libraries (not working with yarn and some npm versions, because the `riot` binary can not be found)


aheinze:
thanks!

Issue #544, opened 2017-05-22
'list' reserved word in PHP 5.5.9


zontarian:
Hi

I've installed cockpit on an Ubuntu with php 5.5.9 and it seems that sometimes, while saving an entry, Cockpit gives an error on `Revisions.php list` definition.

     "PHP message: PHP Parse error:  syntax error, unexpected 'list' (T_LIST), expecting identifier (T_STRING) in /data/www/xxxx/demo/cockpit/modules/Cockpit/Helper/Revisions.php on line 18

It appears `list` cannot be a function/method name.

I've changed two files file:

- cockpit/modules/Collections/Controller/Admin.php : 271
- cockpit/modules/Cockpit/Helper/Revisions.php : 18

At the moment I cannot make a pull request.

zontarian:
I've submitted a pull request

aheinze:
merged

PR #545, opened 2017-05-23
Zontar

zontarian:
The new revisions feature breaks my installation: depending on the PHP parameters it can stop the page from working, or the error can appear on the page, as on my local machine. 
The error is the one reported in the issue #544 
I've changed the name of th method `list` in Revisions to avoid this.

Since I'm also working on an add on for Cockpit-next, I had to correct the .gitignore to exclude the addons dir entirely.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@
 /addons/*/*
 /tests/*
 /node_modules
-
+addons/
 !.gitignore
 !.gitkeep
 !/storage/*/index.html
--- a/modules/Cockpit/Helper/Revisions.php
+++ b/modules/Cockpit/Helper/Revisions.php
@@ -15,7 +15,7 @@ class Revisions extends \Lime\Helper {
         return $this->storage->count("cockpit/revisions", ["_oid" => $id]);
     }
 
-    public function list($id) {
+    public function revlist($id) {
 
         return $this->storage->find("cockpit/revisions", [
             "filter" => ["_oid" => $id],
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -10,7 +10,7 @@
 /addons/*/*
 /tests/*
 /node_modules
-
+addons/
```
piotr-cz: I think better solution is to rename `/addons/.gitkeep` to `/addons/.gitignore` with content

```
*
!.gitignore 
```

modules/Cockpit/Helper/Revisions.php
```diff
@@ -15,7 +15,7 @@ public function count($id) {
         return $this->storage->count("cockpit/revisions", ["_oid" => $id]);
     }
 
-    public function list($id) {
+    public function revlist($id) {
```
aheinze: instead of revlist I would prefer getList

zontarian:
Hi, 
i've done as @aheinze  and @piotr-cz suggested. This is my pull request on a forked github projet, so I don't know if I have done all the steps correctly.

piotr-cz:
Thanks, I've just encountered this bug when saving a collection entry.
Just little hint: next time try to use more descriptive pull request title and you'll get more feedback.

zontarian:
@piotr-cz yes, I come from a small team dev cycle and branches are usually used in a different way. I can see now the use for meaningful branch names, especially for this setup.

piotr-cz:
It's good practice to categorize branches by type (feature/bugfix) but what I meant is the Pull Request Title at the top of this page which reads _Zontar_.

If you would name it something like _Patch for error 500 when saving collection entries_, other developers or the maintainer would more likely open it, review and merge.

zontarian:
Ah ok, i get it, i wasn't aware the pull request pull title could be changed (as I said it was my first on github.. usually we just call out from on desk to the other for the project lead to review and merge) I will request another one in an hour so I'll try. Thanks!

PR #546, opened 2017-05-24
pass along JS options to UIKit html editor and CodeMirror for markdown fields

zontarian:
Added ability to pass configuration options, other than `height`, to the htmleditor used also for markdown. 

Useful for example, to have side by side editing of markdown:  For example, in the options of a collection's field, try writing this: 

    {
        "height": 200, 
        "htmleditor": {
            "mode": "split",
            "maxsplitsize": 800,
            "codemirror": {
                "lineNumbers": true
            }
        }
    }

The `lineNumbers` options in useless in markdown, but just an example here.
It's a bit less useful now than one year ago, since now you can specify the `height` property of the editor as a plain option, but nonetheless can be handy to pass specific options to the underlying UIKit html editor or directly to codemirror.
This is the list of UiKit options (as derived from UIKit' `htmleditor.js` files, is (with defaults)

    {
        iframe       : false,
        mode         : 'split',
        markdown     : false,
        autocomplete : true,
        enablescripts: false,
        height       : 500,
        maxsplitsize : 1000, 
        toolbar      : [ 'bold', 'italic', 'strike', 'link', 'image', 'blockquote', 'listUl', 'listOl' ],
        lblPreview   : 'Preview',
        lblCodeview  : 'HTML',
        lblMarkedview: 'Markdown'
   }

while if you add a `codemirror` key under `htmleditor` key, you can pass options direclty to codemirror as per  [CodeMirror config options](https://codemirror.net/doc/manual.html#api_configuration)

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -1762,6 +1762,15 @@ riot.tag2('field-html', '<textarea ref="input" class="uk-visibility-hidden"></te
 
                 $this.refs.input.value = $this.value;
 
+                //get codemirror options from JSON options and merge in opts
+                if(opts.htmleditoroptions){
+                    var co=opts.htmleditoroptions;
+                    for(var prop in co){
+                        if(!(prop in opts) || opts[prop] === undefined)
+                        opts[prop]=co[prop];
+                    }
+                }
+
                 editor = UIkit.htmleditor(this.refs.input, opts);
                 editor.editor.on('change', function() {
                     $this.$setValue(editor.editor.getValue());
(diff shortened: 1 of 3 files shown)


zontarian:
Any chance of it being merged? I'd really need to customize the HTML editor, to include for example figure and figurecap in the resulting markdown, with a button (and not by typing)..

aheinze:
But you can already do that: https://github.com/agentejo/cockpit/blob/master/modules/Cockpit/assets/components/field-html.tag#L41

zontarian:
Hi Artur,

thanks for pointing that out: but exaclty how can you pass options to 
UIKit html editor? Can you provide an exampe of configuration, via the 
JSON of a field I suppose?

ANd can you forwarard configs also to CodeMirror?
Thanks!

W

Walter Gamba
Amministratore Delegato
IAKTA srl

walter@iakta.it
+39 348 5610576
+39 011 3157422
+skype sakkio

IAKTA SRL - Ingegneria del software
www.iakta.it
via Agostino da Montefeltro, 2
10134 Torino
PIVA CF 08616480011

aheinze:
https://github.com/uikit/uikit/blob/v2/develop/src/js/components/htmleditor.js#L31

example field options:

```
{
   height       : 500,
   maxsplitsize : 800,
   codemirror   : { "lineWrapping": false}
}
```

PR #547, opened 2017-05-24
Fix entries localization

piotr-cz:
At this moment it's not possible to localize entry values (at least in Chrome v58.0.3029.110)
When switching to other language, values stay the same and in database are stored under same keys.

I've tracked the problem down to [assets/lib/riot/riot.bind.js](https://github.com/COCOPi/cockpit/blob/next/assets/lib/riot/riot.bind.js) library.

For some reasons accessing the tag values as dot-separated-keys using current method (eval) the didn't work properly. Switching to non-eval method fixed the issue. Unfortunately I'm not able to explain why.

Steps to reproduce:
1) Add new language in config
2) Create new Collection (Post template) and set some fields as Localized
3) Create new entry, fill out localized fields
4) Switch language (value that are supposed to be localized should clear up)
5) Save (localized values should be stored in database)

### DIFF
--- a/assets/lib/riot/riot.bind.js
+++ b/assets/lib/riot/riot.bind.js
@@ -51,15 +51,9 @@
                     return;
                 }
 
-                var lastObject = tag;
                 var segments = field.split('.');
+                var lastObject = getLastObject(tag, segments);
 
-                // Retrieve last object by segments
-                for (var j = 0; j < segments.length - 1; j++) {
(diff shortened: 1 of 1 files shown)


piotr-cz:
Update: switched to `_.get` and `_.set` methods since lodash is bundled

aheinze:
ah yes, this may work...I need to test it first, but will merge it if everything looks ok. thanks!

PR #593, opened 2017-10-30
Entry / Entries View overrides for Collections and Forms

backtick:
I was trying to make a custom entries view, because I have a lot of columns shown by default, but I couldn't get it to work, until I found some small errors in code.


aheinze:
Thanks for the fix. But you can define in the collection edit view which field should be visible on the entries view. Just toggle the following icon:

image

backtick:
Thanks. Now I feel like a complete idiot...

Well, at least it still fixed something.

PR #613, opened 2017-12-05
Update field-select.tag

ViRuS2k:
{
    "default" : "option1",
    "options" : {
        "option1": "Option 1",
        "option2": "Option 2"
    }
}


piotr-cz:
I think you have to compile assets (`npm run build-components`) and include them in PR

PR #628, opened 2017-12-22
Added full Group management with UI ACL configuration

serjoscha87:
This pull would contain:

- UI Menu for Group management like its already implemented for users. This will render the yaml group configuration obsolete and ensure a much more easy user, group and acl management
- Bulk action when creating a user that enables 

See the screenshots for a better understanding: https://imgur.com/a/MR1Ps


aheinze:
@serjoscha87 thank you for the contribution. but this is a pretty big PR. Do you think it is possible to wrap this code in an add-on for cockpit?

serjoscha87:
I was planing to do so - but cockpit is a really massive peace of code and I had to struggle getting into its structure and logic. At last I think this cannot be easily done with an addon because much stuff, like the menu for example is hardcoded within the cockpit files (bootstrap, admin ...). So I could not add menu entries without modifying your files. Also the Auth/ internal User ACL management is done within the cockpit bootstrap file. I cannot inject dynamic ACLs without changing what once was yours. Basically this is the same with the Settings overview: all entries are hardcoded - I could not add stuff without changing the settings files them self... 

I just took a look at the pull request again an saw that there were some files mistakenly added to the pull (like the "group_old_storms.php" -> this file can be just deleted completely). Further more my less compiler removed a few new lines at the end of the map files resulting in more "files that have been changed" but at last this is nothing we should be concerned of - in fact I only made a few style changes...

aheinze:
yo can dynamically add menu entries by hooking into events also extend ACLs. I don't see any problems regarding that 🤔

serjoscha87:
well, if so, I haven't figured out how to do so yet.

but isn't this the spot where the menu is defined?
https://github.com/agentejo/cockpit/blob/next/modules/Cockpit/views/layouts/app.php

How could I hook into this? 

And what are those @trigger statements for?

aheinze:
e.g. https://github.com/agentejo/cockpit/blob/next/modules/Collections/admin.php#L21-L25

serjoscha87:
Okay, thanks. I'll be looking into it again in order to make a module/addon out of it. Is there a way to ask you, if I get stuck?

serjoscha87:
I made an Addon for the group management, as you desired.
There are a few things that fell aboard by converting my additions into an addon (like the permission view of collections which was able to shrink the group cards down to a smaller layout as soon as there are more then 5 groups [...])

Here it is:
https://github.com/serjoscha87/cockpit_GROUPS

PR #645, opened 2018-01-17
fill select field with default value. minor select field cleanup

gryphonmyers:


modules/Cockpit/assets/components/field-select.tag
```diff
@@ -2,7 +2,7 @@
 
     <select ref="input" class="uk-width-1-1 {opts.cls}" bind="{ opts.bind }">
         <option value=""></option>
-        <option each="{ option,idx in options }" value="{ option }">{ option }</option>
+        <option each="{ option,idx in options }" value="{ option }" selected="{ parent.root.$value === option }">{ option }</option>
```
aheinze: is _selected_ needed here as `this.refs.input.value = this.root.$value;` does that? 🤔

gryphonmyers:
Without this pr, the default doesn't show up on the actual select element -
it remains blank as if no selection has been made

aheinze:
Thanks Gryphon!

Issue #647, opened 2018-01-21
[Proposal] Add a max length option to the text fields


pauloamgomes:
#648  it would be a good enhancement if some fields (e.g. text) can be limited in terms of max length. Having such attribute would improve the editorial experience.

aheinze:
fixed, thanks!

PR #648, opened 2018-01-22
support for maxlength attribute

pauloamgomes:
#647 Adding support for maxlength on text field

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -3069,6 +3069,12 @@ riot.tag2('field-text', '<input ref="input" class="uk-width-1-1" bind="{opts.bin
                 this.refs.input.setAttribute('required', 'required');
             }
 
+            if (opts.attributes) {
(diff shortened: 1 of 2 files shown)


aheinze:
Thank you for the contribution! would be great to have also support for min length, pattern etc.

pauloamgomes:
Yeah, it make sense, the same structure can be used for the field definition, e.g.:

input-attributes-1

or we can have an attributes attribute and just iterate over it and set them:

text-attributes-2

pauloamgomes:
In the meantime updated the PR to support multiple attributes:

cockpit-textfield-attributes

That will result in the proper validations, e.g.:

pattern-fail

aheinze:
sorry, but I prefer the old way ... thanks again!

pauloamgomes:
Ok, but you mean having the maxlength, minlength and pattern on the root definition? Or you don't want at all the change?

aheinze:
On the root definition, the change is great and needed!

PR #711, opened 2018-03-18
Sortable entries items missing the "duplicate" action

mathielen:


aheinze:
Thanks for the contribution but this seems to be fixed already

PR #760, opened 2018-05-08
Improves Defining CLI Constant In bootstrap.php

funkvn:
* Uses identical operator
* Adds a block to if statement (recommended by PSR-2)

Hints comes from Php Inspections (EA Extended) too.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -5,7 +5,9 @@
  */
 define('COCKPIT_START_TIME', microtime(true));
 
-if (!defined('COCKPIT_CLI')) define('COCKPIT_CLI', PHP_SAPI == 'cli');
+if (!defined('COCKPIT_CLI')) {
+    define('COCKPIT_CLI', PHP_SAPI === 'cli');
+}
 
 /*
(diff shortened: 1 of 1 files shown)


funkvn:
Just a minor change to test the workflow. :-)

funkvn:
OK, the change was intendet to merge too.
What do you think about the code?

aheinze:
Thank you for the contribution. But this PR is just a code style preference. Also there is no need for `===` as the value of `PHP_SAPI` is always a string

funkvn:
Thanks for your reply. 

So I will argue the other way around: `PHP_SAPI` is always a string so you do not need a loose comparison. 

`===` is more clear (it says, that a string is expeced, and here it is) and faster (https://stackoverflow.com/questions/2401478/why-is-faster-than-in-php; http://www.phpbench.com/)&mdas;and that's why it is recommended (e. g.) by the PHP Inspections.

To code style again: why not using a common standard as PSR-2? I was really skeptical but over the years this was a big step for code interchange, simplifying the settings settings and easier contribution.  :-)

PR #770, opened 2018-05-29
get app languages in collection api request

casaper:
I don't know if this is of any use, and I admit not reading any contribution guide. I need this for myself, but in case this is useless for upstream, don't it can be ignored or declined...


mpartipilo:
@aheinze I will need this functionality soon. I can put this in an add-in, but don't you think it should be part of core?

PR #787, opened 2018-06-26
[Next] Tags-Field - apply tag field with comma, fix #786

### LINKED ISSUE #786: [Next] Tags-Field - ability to add new tag with comma
I think, it would be more intuitive to apply/separate tags in the tags field with comma, like Wordpress or Youtube and others. The actual behavior is, to apply a new tag with enter and I had to search the issues to find this solution.

I'll send a pull request.

https://github.com/agentejo/cockpit/pull/787

raffaelj:
Now it's possible to press enter or comma to separate tags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 .idea
 .env
 composer.lock
+package-lock.json
 /lib/**/docs/*
 /lib/**/tests/*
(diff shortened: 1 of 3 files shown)


aheinze:
Thank you for the contribution!

PR #800, opened 2018-07-03
fix typo in entry.php

basvdheijden:


basvdheijden:
There is a typo there, you need an extra slash.

aheinze:
Thanks!

PR #812, opened 2018-07-17
Fix method name in SimpleImage

piotr-cz:
Fix method call from `$this->image->output` to `$this->image->toScreen` as   [claviska\SimpleImage](https://github.com/claviska/SimpleImage/tree/3.3.3) doesn't have such method


aheinze:
Thanks!

PR #818, opened 2018-07-19
Sot collection fields by name

piotr-cz:
So it's easier to find field and order is consistent with [documentation](https://getcockpit.com/documentation/reference/fieldtypes)

cockpit-sortfields


aheinze:
Thanks!

PR #823, opened 2018-07-20
Use relative paths to handle component icons

piotr-cz:
Required for https://github.com/agentejo/LayoutComponents/pull/1


aheinze:
I'm not sure about that. this PR would limit the ability to use custom icons 🤔

piotr-cz:
Actually it unlocks the ability to use custom icons by ability to define icon base name (ie.`business.svg`) 
 instead of absolut url which is required at the moment.

aheinze:
I think a compromise would be to introduce a pattern like `icons:code.svg` for icons located in `/assets/app/media/icons/` otherwise handle it as a link.

piotr-cz:
You mean by adding a system path, that would be used as icons directory?
```
$app->path('#icons', COCKPIT_DIR . '/assets/app/media/icons');
```

piotr-cz:
I think adding new path for icons (so it's possile to use icons from outside of `/assets/app/media/icons`) other feature out of scope of this PR, there would have to be changes in any module that uses icons (collection, forms, singletons).

In latest commit I've changed paths handling so it's working with icons in similar way how collection module is:

`App.base('/assets/media/icons' + icon)` -> `'@url('assets:app/media/icons/')' + icon`.

PR #825, opened 2018-07-22
Cp next i18n fixes

raffaelj:
* load app i18n if no user i18n available (login page)
* fixed missing i18n of login, finder and Cancel buttons in notifications


aheinze:
Thanks!

PR #826, opened 2018-07-22
added ability to display singletons in cockpit.menu.aside

raffaelj:
added ability to list singletons in system menu (like collections and forms)


aheinze:
Thanks!

PR #830, opened 2018-07-30
Ability to use geolocation coords

piotr-cz:
Adding ability to use geolocation coords (like `53.55909862554551, 9.998652343749995`) in the places autocomplete input element.

This is handy, when one would like to add location which
- address has changed recently (new street name) and isn't available in autocomplete databse yet
- doesn't have address (place outside of city like in mountains, sea, etc)


aheinze:
Thanks!

PR #833, opened 2018-08-01
plain text mail - pretty markdown without losing links

raffaelj:
The plain text version of sent mails was ugly. strip_tags didn't trim the mail template and it cut out important links like the password reset link.

I added Markdownify to the libraries for a pretty convertion to text mails without losing any information.

I'm not sure, what you think about adding libraries and I wasn't able to stop composer from updating mongodb while requiring Markdownify. I don't use mongodb, so I don't know, if the update could break anything.

The commit titles and descriptions should be self-explanatory.


aheinze:
Why don't you just utilize the custom templates for form emails: https://github.com/agentejo/cockpit/blob/next/modules/Forms/bootstrap.php#L276

And you also don't need to add a custom markdown parser as Cockpit is already shipped with one: https://github.com/agentejo/cockpit/tree/next/lib/vendor/erusev/parsedown

Greets
Artur

raffaelj:
Maybe I'm blind, but I didn't see a way to reverse Parsedown. Markdownify is a HTML parser, that produces markdown.

The custom form mail template is just for the HTML mail part and it's only for forms. To provide better form notifications, I already added a few lines of code to my form validation addon (https://github.com/raffaelj/cockpit_FormValidation/blob/master/bootstrap.php#L94).

The plain text conversion is in Mailer.php, which strips links, too.
https://github.com/agentejo/cockpit/blob/next/lib/Mailer.php#L75

This is, how the plain text version of the password reset mail looks like:

cockpit_password-reset_plain-text
a lot of whitespace and no reset link :cry: 

```
This is a multi-part message in MIME format.
--b1_WWReDGxqnQBvok8jtXcUIPeqmbh8SLpPwjDXSiPPBa0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit



    

       
    
        

            RaffaelTanzt

            Passwort zurÃ¼cksetzen!

        
    

--b1_WWReDGxqnQBvok8jtXcUIPeqmbh8SLpPwjDXSiPPBa0
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 8bit

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html>
    <head>

    </head>   
    <body>
        <div class="container">

            <h1><center>RaffaelTanzt</center></h1>

            <center><a href="http://localhost/cockpit/auth/newpassword?token=MYSECRETTOKEN">Passwort zurÃ¼cksetzen!</a></center>

        </div>
    </body>
</html>


--b1_WWReDGxqnQBvok8jtXcUIPeqmbh8SLpPwjDXSiPPBa0--
```

It would be possible to

* add a second, optional plain-text template
* and replace `strip_tags()` with a function, that strips tags and keeps links as a fallback.

But my solution with Markdownify is global, for all mail notifications and I don't have to care about providing an extra plain-text template.

But I discovered a bug in my PR and I fixed it with my last commit. The converter kept tags like `<center>` by default, why my form notification tests worked and the password reset still had html in the plain text version.
https://github.com/raffaelj/cockpit/commit/3c2c45fc20552b570b48d10aa823dd91c2edffb2

Now it should work like expected.

raffaelj:
No it looks much better:

```
This is a multi-part message in MIME format.
--b1_ORbzkNcA7LPN0RlNjgdqQNB1stFU34bgbohangOH0
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 8bit

# RaffaelTanzt

[Passwort zurÃ¼cksetzen!][1]

 [1]: http://localhost/cockpit/auth/newpassword?token=MYSECRETTOKEN

--b1_ORbzkNcA7LPN0RlNjgdqQNB1stFU34bgbohangOH0
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: 8bit

<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html>
    <head>

    </head>   
    <body>
        <div class="container">

            <h1><center>RaffaelTanzt</center></h1>

            <center><a href="http://localhost/cockpit/auth/newpassword?token=MYSECRETTOKEN">Passwort zurÃ¼cksetzen!</a></center>

        </div>
    </body>
</html>


--b1_ORbzkNcA7LPN0RlNjgdqQNB1stFU34bgbohangOH0--
```

raffaelj:
Since the last commits to add AltMessage and other options in forms.submit.before, all suggestions above are optional through addons or config scripts now. Thanks for adding better options.

PR #842, opened 2018-08-17
avoid redirect to cockpit.start if unauthorized

raffaelj:
If I use `cockpit.start` in config.yaml to skip the dashboard after login and I am not authorized  to see this page (non-admin user), I see a 401 after the login. The button "Get back" leads to `/cockpit`, which redirects me again to the 401.

I added a check, if the user has access before redirecting.

### DIFF
--- a/modules/Cockpit/admin.php
+++ b/modules/Cockpit/admin.php
@@ -82,7 +82,7 @@ $app['app.assets.base'] = $assets;
 
 $app->bind('/', function(){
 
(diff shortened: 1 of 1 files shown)


raffaelj:
I close this - too specicific.

If you need a custom redirect on start, better add your own rules to `config/bootstrap.php`:

```php
$app->on('admin.init', function() {
    $this->bind('/', function(){
        // custom checks for user groups etc.
        $this->reroute('...');
    });
});
```

Alternatively, `cockpit.start` could be user specific instead of global.

PR #844, opened 2018-08-22
added handling of response headers overrides by config

pauloamgomes:
added possibility of overriding response headers based on config, assuming existing defaults when config is not present

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -110,15 +110,6 @@ function cockpit($module = null) {
 
             'filestorage' => [],
 
-            'cors' => [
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks for the contribution. I would like t simplify this.

keep `bootstrap.php` as it is and only update `index.php`:


```
header("Access-Control-Allow-Origin: ".$cockpit->retrieve('config/cors/allowedOrigins', '*'));
.
.
.
```

pauloamgomes:
Ok, updated

aheinze:
Thanks!

PR #849, opened 2018-08-26
added export of entries to RestApi of Forms

### LINKED ISSUE #848: Access Form Entries via API
Is it possible to access the entries of a form through the API? I found nothing in the documentation.
This kind of access should maybe be configurable separately for every Form.

reins-ch:
- would close #848 
- the export method is mostly a copy of the method found in Admin.php, so it gives the same output (JSON) as a click on "Export" in the backend

I'm not sure about the naming conventions. Because I mostly copied the method, I also copied the name. Maybe "entries" is an alternative to "export"

### DIFF
--- a/modules/Forms/Controller/RestApi.php
+++ b/modules/Forms/Controller/RestApi.php
@@ -18,4 +18,17 @@ class RestApi extends \LimeExtra\Controller {
 
         return false;
     }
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks!

Issue #774, opened 2018-05-31
XML Parsing Error in Firefox for field-collectionlink.tag


raffaelj:
I have a
```
XML Parsing Error: not well-formed
Location: http://localhost/cockpit/modules/Collections/assets/field-collectionlink.tag
Line Number 7, Column 59:
```
in Firefox (ESR with script blockers and security settings) and also with a new downloaded FirefoxPortable without plugins. I'm not sure, if some functionality is broken. I still can link collections.

With Opera I don't see this error in the console.

The same with MultipleCollectionLink:

```
XML Parsing Error: not well-formed
Location: http://localhost/cockpit-next/addons/MultipleCollectionLink/assets/field-multiplecollectionlink.tag
Line Number 42, Column 60:

XML Parsing Error: not well-formed
Location: http://localhost/cockpit-next/modules/Collections/assets/field-collectionlink.tag
Line Number 7, Column 59:
```

Often I ignore such messages, but it might be related to #773.

raffaelj:
Update: The error appears only on localhost (Xampp for Windows). Maybe it has something to do with ssl...

With a fresh installed Firefox there is no warning, if it is on a test remote host (Uberspace7). But my Firefox in paranoia mode still gives the warning: `Die Ressource von "https://example.com/cockpit/modules/Collections/assets/field-collectionlink.tag" wurde wegen eines MIME-Typ-Konfliktes blockiert (X-Content-Type-Options: nosniff).`

I have to look further, which settings I changed...

bitmorse:
can confirm I am getting the same error in the console in Firefox 61.0.2 on Arch Linux. All functionality except thumbnails is working fine. Chrome doesn't get this error but thumbnails also don't work there so it doesn't seem to be related.

PR #855, opened 2018-09-11
set .tag files as javascript to avoid xml parsing errors in firefox console

raffaelj:
.tag files were interpreted as xml files (in Firefox), where symbols like "{}" are not allowed. Setting the mime type in .htaccess to application/javascript fixes this behavior.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -39,7 +39,7 @@ Options -Indexes
 
 AddDefaultCharset utf-8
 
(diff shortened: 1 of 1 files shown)


raffaelj:
fixes #774

aheinze:
Thanks!

raffaelj:
It's night-time. You shouldn't work now ;-) Thanks for merging.

Issue #699, opened 2018-03-07
Feature Request: Key Value for multipleselect and select


paulstelzer:
At the moment you can only type in the options as string for multipleselect and select like:

```
{
    "options": "Option 1, Option 2, Option 3"
}
```

Could you add an option to make it key-value like:

```
{
"options": {
"key1": "Option 1",
"key2": "Option 2",
"key3": "Option 3"
}
}
```

So in Cockpit Backend you can check "Option 1", but in json (API output) it is "key1"

woutervddn:
This would simplify things by A LOT!

The current method isn't a generic json format and therefor isn't searchable in vanilla javascript (or php) without building a parser for it.

Though it isn't likely that an option itself would contain a "," we shouldn't assume this.

PR #871, opened 2018-09-26
added value-label functionality to field-select and field-multipleselect

raffaelj:
I added the functionality to use select fields with keys and labels in field-select and field-multipleselect.

Usage:

```json
{
  "options": [
    {
      "value": "a",
      "label": "one"
    },
    {
      "value": "b",
      "label": "two"
    }
  ]
}
```

It is backwards compatible, so `{"options": "a,b,c"}` or `{"options": ["a","b","c"]}` still work.

Only the values are stored - like before.

references:
#699
https://discourse.getcockpit.com/t/select-field-label-for-each-value/145

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -2894,7 +2894,7 @@ riot.tag2('field-location', '<div class="uk-alert" if="{!apiready}"> Loading map
 riot.tag2('field-markdown', '<field-html ref="input" markdown="true" bind="{opts.bind}" height="{opts.height}"></field-html>', '', '', function(opts) {
 });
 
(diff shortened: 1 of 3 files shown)


aheinze:
I will close this because this was implemented in this commit: https://github.com/agentejo/cockpit/commit/047ae6b6256e1eccf5104bec0c627244e84c028a

with the possibility to also define options as `{"a":"one", "b":"two"}`

raffaelj:
Thanks. It still works with my test collection from yesterday for labeled options, but there is a new issue, I eliminated in my closed PR before:

How to reproduce:
* create two select fields with `{"options": "a,b,c"}`
* create or change an entry
* selected options disappear visually, saving works

cockpit_select-issue

Remove https://github.com/agentejo/cockpit/blob/next/modules/Cockpit/assets/components/field-select.tag#L47

from update and add it to mount:

```php
        this.on('mount', function() {
            this.refs.input.value = this.root.$value;
            this.update();
        });
```

PR #875, opened 2018-09-30
catch console error of jsoneditor if field object contains invalid json

raffaelj:
When the json in an object field is invalid (or empty string), each text
change fires a very long error message in the browser console.

The code editor still has a red "x", when it detects invalid content, so
there is no need for a custom console output (`catch (e) {}`).

`Error in onChange callback:  Error: Parse error on line 1: ...`

I found the solution here:
https://github.com/josdejong/jsoneditor/issues/413


raffaelj:
Thanks for fixing it in https://github.com/agentejo/cockpit/commit/736d150a06c569ba75453da49ee0425eeff08d9b

Issue #873, opened 2018-09-28
Assets rerouted Instatlation Issue


PiIsFour:
Hello,

i have some trouble installing cockpit. Probablly something i missed, but i would be really thankfull if someone can point me in the right direction. Don't even know how to diagnose it further.

The installation locally worked fine, but now i tried to upload it to my webserver (strato). The install went without error but the login page is brocken. It seems to me some problem with mod rewrite because all the assets (like jquery) get reroutet with a 302 to /api/auth/login.

I have installed it under a subdomain, if that makes a difference.

Thank you for your time and help.

raffaelj:
You don't need to change the .htaccess.

Create a file `defines.php` in your cockpit root directory with

```php
define('COCKPIT_BASE_URL', '/' . basename(__DIR__));
define('COCKPIT_BASE_ROUTE', '/' . basename(__DIR__));
define('COCKPIT_DOCS_ROOT', dirname(__DIR__));
```

And I wrote an install script for strato, because it was very annoying to have such a limited console:
https://github.com/raffaelj/cockpit-scripts/blob/master/install-cockpit-on-strato/install.sh

PiIsFour:
Thank you @raffaelj,

that sounds good. Will try it tomorrow.

PiIsFour:
Finally got time to do it and it worked. Just added the defines.php and put in my pathnames.

Thanks again @raffaelj

PiIsFour:
Sorry, but maybe i closed this to early. Should i open a new one?

Last time i only checked that the frontend of Cockpit worked on my server. But now i filled in my data and tried to fetch it. but now i get a 404 on the api request.

i can access the frontend at: beta.MYDOMAIN/cms
so shouldn't i get the api with: beta.MYDOMAIN/cms/api/collections/get/test
that is what worked locally.

I'm really sorry for asking all this questions.

raffaelj:
Sorry, I used cockpit as a library to have a backend for a simple php frontend. I didn't notice, that the api doesn't work. It worked on my localhost and apparently I never checked, if the api works on the server.

I tested it again. The backend and frontend work. When trying to call the api through `https://example.com/admin/api/...`, I got redirected to the login page. If calling `https://www.example.com/admin/api/...`, I get a `{"error": "404", "message":"File not found"}`.

There was some weird config, I had to add to load cockpit as library ([found it here](https://github.com/agentejo/cockpit/issues/371#issuecomment-162849916)):

```php
[...]

PR #878, opened 2018-10-03
moved order of defined constants to take care of COCKPIT_BASE_URL for api request

raffaelj:
If COCKPIT_BASE_URL was defined in defines.php for using Cockpit in a subfolder, the backend worked and the api calls redirected to login.

That happened, because COCKPIT_API_REQUEST was never set to '1' after comparing the auto-guessed $COCKPIT_BASE_URL instead of defined COCKPIT_BASE_URL.

Tested by me and @PilsFour an a shared host on strato.

fixes: #873
references: https://discourse.getcockpit.com/t/cant-access-api/221

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -49,12 +49,12 @@ $COCKPIT_BASE_ROUTE  = $COCKPIT_BASE_URL;
  * SYSTEM DEFINES
  */
 if (!defined('COCKPIT_ADMIN'))                  define('COCKPIT_ADMIN'          , 0);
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks!

fais3000:
@raffaelj Interestingly for me api in subfolder still doesn't work with above fix. I still get 404 on api endpoints, while frontend works.

raffaelj:
@fais3000 Please open a new issue and add some descriptions. Otherwise it's hard to help you.
Which Cockpit version do you use? Is your PHP version >7.0? Maybe it is a typo in your request...

PR #880, opened 2018-10-04
made $data and $frm in custom form mail template available, fixes #879

### LINKED ISSUE #879: forms - mail template - Undefined variable: data and $frm not available
When the custom form mail template is rendered, $data is not available and also not the form name.

https://github.com/agentejo/cockpit/blob/next/modules/Forms/bootstrap.php#L280

`$body = $this->app->renderer->file($template, $data, false);`  should be
`$body = $this->app->renderer->file($template, ['data' => $data, 'frm' => $frm], false);`

raffaelj:
fixes #879


aheinze:
Thanks!

Issue #863, opened 2018-09-18
New settings section for addons with enable/disable/reorder option


pauloamgomes:
Think would be interesting if we can have in the settings page a new section for addons, that just list them with some info and provide option to enable/disable/reorder (without removing them from the filesystem). 
So that way only the enable addons/modules would be loaded during the bootstrap, I can see that the loadModules() is loading everything from the addons and modules folder.
That would also permit to set dependencies (load addon x after y is initialized)

aheinze:
I think it is a good idea to have an overview of installed addons but the other suggestions would complicate stuff. Need to think about a bigger picture here (addons installer etc).

Thanks for you suggestions

pauloamgomes:
Yep, I understand your point, at least we can start only with the overview, simple list/grid of addons with name and info.

The other part would require more investigation, it can be similar to a market place of addons, where you can install/uninstall directly from the UI (e.g. similar to wordpress)

raffaelj:
I like the idea. Some suggestions:

Allow deactivating core modules, too (if I don't need Singletons and Forms)

Addons could have a json file or an addon.info trigger with infos about it
  * [example of Nextcloud - markdown addon](https://github.com/icewind1991/files_markdown/blob/master/appinfo/info.xml) --> with xml files
  * could also contain info about last testet cockpit version...

@jakerb started to write a marketplace addon 10 months ago. I'm not sure, if it works and if it was exactly for installing addons, but it looks so from skimming the code.
  * https://github.com/jakerb/cockpit-module-marketplace
  * https://github.com/jakerb/cockpit-marketplace

But an actual installer seems to be a lot of work. Disabling or enabling modules through config.yaml would be enough to start with.

```yaml
modules:
    singletons  : false
addons:
    Groups      : true
    UniqueSlugs : true
    Feed        : false
```

A GUI for addon management could be an extra addon.

Some more questions, that came in my mind:

Should addons be enabled by default?
Is it better to deactivate paths or addon names?

raffaelj:
I wrote a base to start with: https://github.com/raffaelj/cockpit/tree/cp-next_deactivate-modules-addons

I didn't change a lot of code: https://github.com/raffaelj/cockpit/commit/0191e49160b3d02dd4e7cdca7f81d5fca98aae3a

add modules to `config.yaml`

```
active.modules:
    singletons  : false
    forms       : true
    feed        : false
```

overview: `url/to/cockpit/addons/index` or via settings menu

[...]

PR #883, opened 2018-10-06
Feature 863 extensions

pauloamgomes:
PR contains minimal changes in core that can provide the functionality requested on #863


pauloamgomes:
@aheinze , any idea if you go ahead with that solution, or you are thinking about something different? The current solution seems simple enough and doesn't cause any regression.

aheinze:
Hi Paulo,

can we change the settings to:

```
modules:
    disabled: [...listOfModules]
```

I think that it makes more sense to define the disabled modules instead of naming the enabled.
What do you think?

Greets
Artur

mpartipilo:
I like how GatsbyJS lets you specify exactly which plugins to use and pass an options object to the plugin instance. Even multiple instances of a plugin can be used with different options.

pauloamgomes:
@aheinze yeah, that can also work and means we don't need to specify the core modules in the code, my initial thinking was based on the idea of enabling a module would require always a user action in the configuration (to avoid for example an addon causing  an error until is properly installed)

pauloamgomes:
Think now is more cleaner, but we need a different key as modules is rewritten:

```yaml
modules.disabled:
  - Regions
  - Forms
  - ...
```

aheinze:
Thanks!

PR #899, opened 2018-10-20
added missing acl actions

raffaelj:
thinking about acls and rewriting the GROUPS addon to list acls generic... so acls must be specified correctly

"singletons - manage" has no effect, but all other modules have a manage action already


aheinze:
Thanks!

Issue #900, opened 2018-10-21
colortag field options have no effect


raffaelj:
The colortag field always uses the default colors. Adding options like `{"colors": ["#aaaaaa", "#00ff00", "#00ffff"]}` has no effect.

PR #901, opened 2018-10-21
mount opts in nested riot tags (fields: tags, colortag)

raffaelj:
fixes: #900

related: https://discourse.getcockpit.com/t/reuse-core-fields-pass-opts-to-nested-riot-tags-with-inconsistent-mount-update-events/348


aheinze:
Thanks!

PR #909, opened 2018-11-05
added renderer for set field

raffaelj:
See this discussion for infos and usage

https://discourse.getcockpit.com/t/renderer-for-set-field-in-backend/372/4


raffaelj:
I added a fallback to prevent empty fields for existent setups, that don't have `"display":true` options and I added a fix to display set fields in the sortable view, too.

Updated screenshots: https://discourse.getcockpit.com/t/renderer-for-set-field-in-backend/372/6?u=raffaelj

Now it should work everywhere.

raffaelj:
Thanks for adding meta info and a simple set renderer in:

https://github.com/agentejo/cockpit/commit/907c1de5ba92f7bbab25635ad20d8b1d5d43a099
https://github.com/agentejo/cockpit/commit/dfc929f4fee57ac59c50b7d841276332c961d5aa

anthonyivol:
Is there a way to display an image on reorder mode ? 

This is my config :

`{
  "display": "image.path",
  "fields": [
    {
      "type": "set",
      "name": "image",
      "label": "Image",
      "options": {
        "fields": [
          {
            "type": "select",
            "name": "size",
            "label": "Taille",
            "require": true,
            "options": {
              "options": "XS, SM, MD, LG"
            }
          },
          {
            "type": "asset",
            "name": "image",
            "label": "Image"
          }
        ]
      }
    },
    {
      "type": "gallery",
      "name": "gallery_xs",
      "label": "Petite gallerie"
    },
    {
      "type": "gallery",
      "name": "gallery_l",
      "label": "Grande gallerie"
    }
  ]
}`

image

As the path is truncate and with a hash at the begining it doesn't help anything on reordering.

I didn't find anything in the doc as the only exemple is : `"display" : null`

I tried with some hazardous things like :
`"display": "",`
and
`"display": true` in all subfield
But this does not work.

I maybe have missed something, it looks like a simple behaviors, any idea ?

raffaelj:
@anthonyivol My idea of the set renderer wasn't implemented. You can use a syntax like this: `"display": "{title} / {from} - {to}"`, but it doesn't work with nested values and it's meant to return a string.

Or write your own renderer for custom needs. I wrote a mini tutorial how to use my solution above as a custom renderer here: https://github.com/raffaelj/cockpit-scripts/blob/master/custom-renderers/README.md

anthonyivol:
@raffaelj thank you ! 
I didn't know this feature and I didn't find something about that in the doc. Anyway that work.

anthonyivol:
@raffaelj I maybe missed something, it works as expected on the admin panel, but my front-office now throw errors as the "admin.init" hook is called on API calls: 
with (config/bootstrap.php) : 
```
$app->on('admin.init', function(){
      $this->helper('admin')->addAssets('/custom/reordering.js');
});
```
I get :
```
Call to a member function addAssets() on null
```
and with:
```
if( $this->helper('admin') != null){
    ...
}
```
I get :
```
<b>Notice</b>:  Undefined index: admin in <b>/xxxxxxxx/lib/Lime/App.php</b> on line <b>1274</b><br />
```
Any idea ?

raffaelj:
@anthonyivol Try

```
if (COCKPIT_ADMIN && !COCKPIT_API_REQUEST) {
    $app->on('admin.init', function(){
      $this->helper('admin')->addAssets('/custom/reordering.js');
    });
}
```

anthonyivol:
Thanks !  That works

Issue #912, opened 2018-11-06
Enhance field.layout.components event with field options


pauloamgomes:
The event **field.layout.components** on the layout field provides access to the component list and therefore modifications on it, however it can be more flexible if we have also access to the field definition. For example, let's imagine that we have a layout field in a collection and we want only to have included a set of components! If we can set something in the field options like:

```
{
  "enabled": [
    "text",
    "image",
    "..."
  ]
}
```

so if in the field we pass the opts:

`App.trigger('field.layout.components', {components:this.components, opts:opts});`

we can handle the event and implement further logic as required.

PR #913 created

PR #913, opened 2018-11-06
passing opts to field.layout.components event

pauloamgomes:
Should provide what was requested on #912


pauloamgomes:
Hey @aheinze , think that one is safe and it will be quite handy for the scenarios that described (In my case want to define a list of allowed components in the field options)

aheinze:
Thanks!

Issue #918, opened 2018-11-09
Focal Point not working, if not the full url path to the image


poldixd:
I have an vertical image and I set the focal point in the asset interface to the top, because the face is in the top section of the image.

I need to create a square thumbnail image. To generate the thumbnail I use `https://example.com/api/cockpit/image` and the direct output `o=1`.

The collection returns a relative path to the image like `/storage/uploads/2018/10/04/image.jpg`. So this must be the url to the thumb (of course... src url encoded):

`https://example.com/api/cockpit/image?token=XXX&q=80&w=200&h=200&m=thumbnail&src=/storage/uploads/2018/10/04/image.jpg&o=1`

But this url does not affected the focal point. If i add the full path of cockpit to the `src`, it works:

`https://example.com/api/cockpit/image?token=XXX&q=80&w=200&h=200&m=thumbnail&src=https://example.com/storage/uploads/2018/10/04/image.jpg&o=1`

It seems, that cockpit does not recognize that the image is an asset:

https://github.com/agentejo/cockpit/blob/2d79c26856799f179ef91d776165d1867e430b35/modules/Cockpit/bootstrap.php#L125

With relative path to the asset `$fp` is always `center`

https://github.com/agentejo/cockpit/blob/2d79c26856799f179ef91d776165d1867e430b35/modules/Cockpit/bootstrap.php#L227

aheinze:
if you're using assets, then please use the asset `_id` as `src` e.g:

```
https://example.com/api/cockpit/image?token=XXX&q=80&w=200&h=200&m=thumbnail&src=5a00e80002ac8doc1362529026&o=1
```

cockpit then also get the focal point settings

pauloamgomes:
> if you're using assets, then please use the asset `_id` as `src` e.g:
> 
> ```
> https://example.com/api/cockpit/image?token=XXX&q=80&w=200&h=200&m=thumbnail&src=5a00e80002ac8doc1362529026&o=1
> ```
> cockpit then also get the focal point settings

That's great @aheinze , was not aware that fp would be handled by the thumbnail function automatically.
Updated the image styles addon so when generating styles for assets they will take the fp in consideration. 

Think the asset field could be improved by providing a quick way to access the asset details (and set there the focal point), or at least provide some events that can handled:

<img width="325" alt="image" src="https://user-images.githubusercontent.com/102261/48293893-10460480-e479-11e8-9a68-4f55a88b03e0.png">

[...]

PR #921, opened 2018-11-10
assets fields refactoring

pauloamgomes:
Related to discussion on #918 

Extracted cp-asset from cp-assets.
Added edit asset functionality from the asset field

### DIFF
--- a/modules/Cockpit/assets/cockpit.js
+++ b/modules/Cockpit/assets/cockpit.js
@@ -97,6 +97,29 @@
 
         assets: {
 
+            edit: function(callback, el){
+
+                var dialog = UIkit.modal.dialog([
+                    '<div>',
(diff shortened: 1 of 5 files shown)


aheinze:
Hi @pauloamgomes !

Thank you for the contribution. Unfortunately I did already start with an implementation that is not that far away from yours ;-)

https://github.com/agentejo/cockpit/commit/38c32d88293dcfec362f871e4275286147081566

pauloamgomes:
@aheinze , yeah, that's fine, I'm just concerned about having the functionality there :)

just one doubt that I found during the implementation since the fp values are saved in the collection, when you edit an entry (at least the way I was doing), you'll have a different fp than the one configured directly in the asset

PR #924, opened 2018-11-13
Add CORS for all uploaded assets

yuhe00:


aheinze:
this is a very case specific and should be implemented individually. Btw Cockpit gives config control over cors values: https://github.com/agentejo/cockpit/blob/next/index.php#L22-L29

Issue #930, opened 2018-11-19
Move trigger event on settings info view


pauloamgomes:
Right now its possible to hook into the settings page and provide further details, however the position of the trigger events in the dom will not permit make use of the sidebar navigation. My suggestion is to move both trigger commands to inside the corresponding containers:

**main contents:**
```html
                </div>

            @trigger("cockpit.settings.infopage.main")
            </div>
```
**sidebar:**
```html
            <ul class="uk-nav uk-nav-side" data-uk-switcher="connect:'#settings-info'">
                <li><a href="#SYSTEM">System</a></li>
                <li><a href="#PHP">PHP</a></li>
                @trigger("cockpit.settings.infopage.aside")
            </ul>
```

PR #931, opened 2018-11-19
issue #930 - move trigger events inside dom to facilitate usage of si…

pauloamgomes:
Change position of trigger events so addons can add items to the sidebar navigation.

### DIFF
--- a/modules/Cockpit/views/settings/info.php
+++ b/modules/Cockpit/views/settings/info.php
@@ -106,10 +106,8 @@
                     </table>
                 </div>
 
-            </div>
-
-
             @trigger("cockpit.settings.infopage.main")
+            </div>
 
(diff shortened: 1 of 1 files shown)


pauloamgomes:
Hey @aheinze , believe this is one is also quite harmless and it will permit other addons to make use of system page to display specific status.

aheinze:
I don't to move them to be honest. this would prevent other possible use-cases. as a compromise: what about introducing `@trigger("cockpit.settings.infopage.aside.menu")` and `@trigger("cockpit.settings.infopage.main.menu")` ?

pauloamgomes:
Yeah, understand, that make more sense.

aheinze:
Thanks!

PR #944, opened 2018-12-02
added upload max size for assets

pauloamgomes:
Based on discussion from https://discourse.getcockpit.com/t/limit-asset-upload-size/411/3, basically just add another check (if the configuration is not defined it works as before).


aheinze:
Thanks!

PR #946, opened 2018-12-03
Fix $gte & $lte working like $gt & $lt

DavidB137:
`$gte` should mean *greater than or equal to*, but now it is just alias of `$gt`. So it doesn't select equal values. The same problem is with `$lte` and `$lt`.


aheinze:
Thanks!

PR #964, opened 2019-01-08
support for default attribute

pauloamgomes:
Support for default on fields select, text, textarea and rating - Issue #965


pauloamgomes:
Reopen as changes on setting defaults seems to work with custom components.

pauloamgomes:
@aheinze , when you can, kindly confirm if that make sense, I'm working in an implementation that uses custom components and having defaults (checkboxes, dropdowns, etc..) is mandatory. Tbh not sure why defaults work on collections but not on custom components, but forcing the verification and update on each field seems to solve it.

aheinze:
I need some time to figure this out. A generic solution for ALL fields would be cool. maybe extend/improve the `cp-field` component 🤔

aheinze:
awesome find! thank you!

PR #974, opened 2019-01-22
Remove double space in string to make it searchable

Crease29:
I had this error message coming up in my installation and I wanted to search for it but no occurence was found. Now it's possible to copy this error message and search for it in the code.


aheinze:
Thanks!

Issue #975, opened 2019-01-24
Default Components: Uncaught TypeError: Cannot read property 'default' of undefined


mortensassi:
I can't add a default layout component, e.g. Image, Headline, Gallery – Only Section or Grid. 
When i click on one of these components i get this error:

```
Uncaught TypeError: Cannot read property 'default' of undefined
    at components.js?ver=1548345771:2832
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (components.js?ver=1548345771:2831)
    at Object.handleEvent (riot.js?ver=1548345771:1456)
(anonymous) @ components.js?ver=1548345771:2832
(anonymous) @ components.js?ver=1548345771:2831
handleEvent @ riot.js?ver=1548345771:1456
```

This is my `docker-compose.yml`

```yml
version: "3"

services:
  cms:
    image: "agentejo/cockpit"
    volumes:
      - ./www:/var/www/html
    ports:
      - '8420:80'
```

If i leave out the `volumes` part, i don't get this error which makes it hard to understand what causes the problem. What's the issue here?

pauloamgomes:
@aheinze believe that error was introduced with my PR for using defaults on custom components, provided a new fix  #976  that adds a missing validation

mortensassi:
Tested your changes – It works! Thanks for the quick fix

PR #976, opened 2019-01-24
mising validation for defaults

pauloamgomes:
#975  Missing validation when checking for defaults.. it seems it was not working with core components but only with custom ones :S


aheinze:
Thanks!

PR #985, opened 2019-01-30
Remove always present scrollbar, fix build documentation

Crease29:
There is a scrollbar that is always visible, even if there is no need to scroll.
I also adjusted the information in the README.md that you only need to run `npm run build` when JS components were modified. It's also necessary when you change LESS files.


aheinze:
I would rather adjust this: https://github.com/agentejo/cockpit/blob/b0b15ffc9f7453fee8fc47a466f26fbad8457865/assets/app/css/less/uikit/base.less#L59

aheinze:
Should be fixed with this commit https://github.com/agentejo/cockpit/commit/ca1fd8d6fc1868f0252a20127bbeed477ffcb2f5

PR #997, opened 2019-02-05
fix(image-thumbnails): re-enable filter 'edge detect' (bc)

jankal:
The method `edge detect` does not exist. This PR fixes all future method names with a space in between.
```
Warning: call_user_func_array() expects parameter 1 to be a valid callback, class 'claviska\SimpleImage' does not have a method 'edge detect' in /var/www/modules/Cockpit/bootstrap.php on line 271
```

See for proof of concept: https://3v4l.org/ZcJOG

### DIFF
--- a/modules/Cockpit/bootstrap.php
+++ b/modules/Cockpit/bootstrap.php
@@ -262,9 +262,18 @@ $this->module('cockpit')->extend([
 
                 // Apply single filter
                 foreach ($_filters as $f) {
+                    $filterName = $f;
+                    // fix filters with spaces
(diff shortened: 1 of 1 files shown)


aheinze:
I close this as it was just a naming typo, no need for special handling/detection

jankal:
@aheinze I really don't know wether this hase in fact ever worked. When it once has, we need to keep bc here. Hence the PR. Else: just rename.

aheinze:
@jankal it was just a typo, so nothing to worry here. thanks for the hint!

jankal:
@aheinze okay. Thanks for the information.

Issue #661, opened 2018-02-06
docs: List of Events for webhooks


TechnicallyJoe:
Hi,

I'm trying to create a webhook, that will shoot whenever a specific collection has been changed (e.g a collection of posts, and whenever a new post is created, an exisiting one is updated or something is deleted then fire webhook)

However the documentation is horrendously lacking and i can't find out what to write in the 'add events' portion of creating a webhook.

I've tried simply writing whats suggested at the bottom of the docs page. 
    collections.save.after
    collections.remove.after
    regions.update
    regions.remove

However it doesn't appear to work. Also is there a more comprehensive list of events available?

Any helps appreciated!

Docs: https://getcockpit.com/documentation/api/webhooks

tomdupont:
Hi,

I tested it successful with this setup:
- My collection name is "works"
- Make sure the webhook is active/on
- Make sure you press the enter key to submit your event to the event list (this is a thing I missed)

![screen shot 2018-02-06 at 16 13 02](https://user-images.githubusercontent.com/15290357/35866947-d2a30898-0b58-11e8-910c-681ef89ead45.png)

tomdupont:
Hi TechnicallyJoe,

I did have the same problem, it's indeed not obvious. I will look if I have the javascript knowledge to add a button and make a pull request.

Here is a list of all the available events **for collections** I found in the latest release. Replace the ones with name, with a specific collection name you have to target only that collection.

- collections.createcollection
- collections.updatecollection
- collections.updatecollection.name
- collections.removecollection
- collections.removecollection.name
- collections.find.before
- collections.find.before.name
- collections.find.after
- collections.find.after.name
- collections.save.before
- collections.save.before.name
- collections.save.after
- collections.save.after.name
- collections.remove.before
- collections.remove.before.name
- collections.remove.after
- collections.remove.after.name

There are many more, but I'm afraid there isn't a documented list of them anywhere yet.

TechnicallyJoe:
I'll upvote the heck out of your PR, if you'll toss what you have right now, into the wiki. Thanks! - Much appreciated. 

I'll revise the issue to include asking for documentation on this. Then you can reference this issue with the PR, if you feel up for it! :)

bassauer-storms-media:
what about sending the data of a fresh created collection entry along to the hook destination? Does anyone know how to do so?

raffaelj:
[...]

PR #1004, opened 2019-02-10
Feature/webhook triggers autocomplete

gobeli:
I was a bit frustrated to figure out which triggers exist for webhooks, so I created an autocomplete in the webhook panel:

autocomplete

I wasn't exactly sure where to put the list (I used the one from #661). Maybe it would be cool to even load it dynamically with the collection/singelton/form names... 

If you are interested to add this feature I would appreciate feedback, which I would be happy to implement.


aheinze:
Thanks for the contribution!

pauloamgomes:
@aheinze , think it may be a problem for triggers that have a dynamic value like the collections.find.after{$name}:

image

I'm not able to change it anymore

aheinze:
@pauloamgomes should be fixed now with this commit: https://github.com/agentejo/cockpit/commit/49f6b1d50713839619e519e71f8e8d1e5f9ebda7

gobeli:
I think it would be cool if the name placeholders would get filled out dynamically with the collection/singelton/form names unfortunately I don't know the project well enough to implement that...

PR #1008, opened 2019-02-10
Boolean field fixes

### LINKED ISSUE #1006: boolean field - renderer doesn't detect string "0" as false
I'm not sure, if there are internal checks for all normal edge cases, but I currently have some data with ones and zeros instead of true/false. The renderer displays "On" for all values, because my zero is a string, which returns true.

The reason: I work on implenting MySQL (InnoDB) as an addon. MySQL has no boolean field and I have to work with tinyint.

### LINKED ISSUE #1007: boolean field - toggle doesn't display a change when using tab and then space key
Steps to reproduce:
* click in a text field
* press tab key to jump to the next field (a boolean field)
* press space to change the state

The state changes, it displays "Off", but the pretty slider stays in the old position.

raffaelj:
fixes #1006
fixes #1007

### DIFF
--- a/assets/app/js/app.utils.js
+++ b/assets/app/js/app.utils.js
@@ -275,6 +275,7 @@
     };
 
     App.Utils.renderer.boolean = function(v) {
+        v = typeof v == 'string' ? parseInt(v) : v;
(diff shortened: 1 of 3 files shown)


aheinze:
Thanks!

PR #1013, opened 2019-02-14
Removed title and description from singleton svg icon

peterbrinck:
I noticed when hovering over the singleton icon, it would show a title.

So I have removed the title and description from the svg icon.

singleton_icon_title


aheinze:
Thanks!

PR #1016, opened 2019-02-15
Fixes collection-link display in Collection view

danailov-z:
It uses same ID for more than one record and crashes visualization:

image

Changed to use ClassName instead


danailov-z:
Fixed with https://github.com/agentejo/cockpit/commit/ab718dc5bf4fd4328149520ae00f956e9094d9f5

Issue #1014, opened 2019-02-15
Catastrophic failure if bootstrap.php doesn't exist in an Addon directory


dlolsen:
### Source of the Problem
The `registerModule` function on [line 1281 of lib/Lime/App.php](https://github.com/agentejo/cockpit/blob/243a39eb7db957f8ee41ca68b25839376bd56ecd/lib/Lime/App.php#L1281) doesn't check if `bootstrap.php` exists in the `addonNameDirectory` before assigning it as the bootfile for that specific module. If `bootstrap.php` doesn't exits the whole app fails.

### History
This failure happened to me when I was trying to install the *UniqueSlugs* addon. Since I was running a Docker container for Cockpit, I didn't know exactly the best way to interact with the filesystem (Docker is still new for me). So I thought I'd just use the _Finder_ in Cockpit to create the `UniqueSlugs` directory inside of `addons`, and then planned to create and copy the contents of the `bootstrap.php` and `version.txt` files... I never made it to adding the files because of the following catastrophic errors.

```
Warning: require(/var/www/html/addons/UniqueSlugs/bootstrap.php): failed to open stream: No such file or directory in /var/www/html/lib/Lime/App.php on line 1346

Fatal error: require(): Failed opening required '/var/www/html/addons/UniqueSlugs/bootstrap.php' (include_path='.:/usr/local/lib/php') in /var/www/html/lib/Lime/App.php on line 1346
```
### <a id="steps-to-reproduce"></a>Steps to reproduce:

1. Create new directory inside `addons` via Cockpit's _Finder_ (in my case it was _UniqueSlugs_)
2. An error alert shows up: *"Something went wrong."*
3. Refresh the page => catastrophic failure

### Suggested Solution
Update line 1281
FROM:
```
$module->_bootfile = "{$dir}/bootstrap.php";
```

TO:

```
if(!file_exists("{$dir}/bootstrap.php")){return false;}
$module->_bootfile = "{$dir}/bootstrap.php";
```

Also, update line 1312 (<--line number prior to above changes)

FROM:

```
$this->registerModule($name, $module->getRealPath());
```

TO:

```
if ( $this->registerModule($name, $module->getRealPath()) === false ) continue;
```

### Conclusion
I suppose you may also want to display an error to the user indicating that there is a _bootstrap.php_ file expected.

FYI, this is the first real issue I've ever submitted on GitHub, and I've also never forked a repo and issued a pull request so I wasn't quite ready for that…

[...]

PR #1019, opened 2019-02-15
Prevent Catastrophic error by making registerModule check if bootstrap.php exists

dlolsen:
Prevent Catastrophic failure in `lib/Lime/App.php` by making registerModule check if bootstrap.php exists. See issue #1014 

### Proposed Fix
- make registerModule return false if no bootstrap.php file exits:
line 1281
```
            /* return false if no bootstrap.php file exists */
            if(!file_exists("{$dir}/bootstrap.php")){return false;}
            $module->_bootfile = "{$dir}/bootstrap.php";
```

- make loadModules skip the dir if registerModule returns false
line 1312 (<-- prior to above fix line 1315 after)
```
                    /* skip loading the module if registerModule returns false */
                    if ( $this->registerModule($name, $module->getRealPath()) === false ) continue;

```

### DIFF
--- a/lib/Lime/App.php
+++ b/lib/Lime/App.php
@@ -1278,6 +1278,9 @@ class App implements \ArrayAccess {
             $module = new Module($this);
 
             $module->_dir      = $dir;
+            
+            /* return false if no bootstrap.php file exists */
+            if(!file_exists("{$dir}/bootstrap.php")){return false;}
             $module->_bootfile = "{$dir}/bootstrap.php";
 
             $this->path($name, $dir);
@@ -1309,7 +1312,8 @@ class App implements \ArrayAccess {
 
                     if ($disabled && in_array($name, $disabled)) continue;
 
-                    $this->registerModule($name, $module->getRealPath());
+                    /* skip loading the module if registerModule returns false */
+                    if ( $this->registerModule($name, $module->getRealPath()) === false ) continue;
 
                     $modules[] = strtolower($module);
                 }


dlolsen:
One last thing, I was unsure what you expected in terms of "tests" as mentioned in CONTRIBUTING.md.

I tested it on my test production Docker container by making the above changes in `lib/Lime/App.php`. I did this by running `sudo docker exec -t -i my-container-name /bin/bash` to get a terminal inside the container, then I just navigated to `lib/Lime/App.php` and made the changes.

I then tried to repeat the error (see [#1014 Steps to Reproduce](https://github.com/agentejo/cockpit/issues/1014#steps-to-reproduce)), and the error would not repeat.

Crease29:
I would not suggest to merge this. An addon needs a `bootstrap.php` per convention. Instead of `if(!file_exists("{$dir}/bootstrap.php")){return false;}` I would throw an `RuntimeException` but not just ignore the error.

dlolsen:
Thanks for looking this over, but I have to disagree that my solution _"ignore[s] the error."_ I believe the underlying fault is in `registerModule` where it **assumes** that the `bootstrap.php` file exists by assigning it:  `$module->_bootfile = "{$dir}/bootstrap.php";`

Since Cockpit allows `Finder` to alter the `addons` directory, this implicitly creates the case where a directory may exist (briefly or otherwise) which has no `bootstrap.php` file. When this happens the whole system breaks and the user is prevented from even creating the `bootstrap.php` file.

Therefore, precisely **because an addon needs to have a** `bootstrap.php` file, my solution proposes to prevent loading a module (addon) which doesn't have one. You'll notice the second part of my solution is to `continue` past the faulty addon in `loadModules` and not actually load it.

For now I believe it is more correct to prevent loading a module without a `bootstrap.php` because it prevents catastrophic failure & allows the opportunity to add the `bootstrap.php` file.


**Side Note:** There doesn't even appear to be a `RuntimeException` (or even, when I tested, an `Exception`... or `Error`) class available in `Lime`  (see image below).

screen shot 2019-02-21 at 10 24 14 am

pauloamgomes:
I agree with the idea if the addon doesn't fulfill the requirements it should not be loaded, that would avoid potential mistakes in a production environment (just creating an empty directory is enough to break cockpit).
Secondly, think the Finder should be limited in terms of action to the storage folder (or probably better if just the uploads folder), having the possibility to access PHP files from the UI is always a bad idea.
That can be already achieved in the configuration:

```yaml
$vars:
  finder.path: /storage/uploads
```

or deny the access to non admin users using:

```yaml
  role:
    cockpit:
      finder: false
```

aheinze:
Thank you for the contribution. But I have to say that this won't get merged.

Reason:

the usage of `file_exists`. `file_exists` comes with performance costs, which I don't want to pay just to check a convention. Doing some tests on my dev env, the req/s dropped from 1536 to 1289 by introducing the check for the bootstrap file.

Crease29:
> **Side Note:** There doesn't even appear to be a `RuntimeException` (or even, when I tested, an `Exception`... or `Error`) class available in `Lime` (see image below).

@dlolsen: You would have to `throw new \RuntimeException('Error message');` (note the leading backslash to use the root namespace).

I understand your case that creating a directory could cause a fatal error and the system won't be usable after that.

How about this compromise: Do it almost like you proposed it here when the config value of `debug` is `false`. Throw a RuntimeException when `debug: true`. I'm asking for that exception from the view of a developer. So if I as a developer start creating an addon and I forget to create the bootstrap.php I want to know what's going wrong and why my addon isn't working. That would also go well with @aheinze talking about the performance issue, which really would be a huge one. So in debug mode the check can be done to give the developer a meaningful hint. In production mode where requests/second are an important value and `debug` should be set to `false`, there won't be an error.

So I wrote "do it *almost* like you proposed". What I meant with that is to try/catch the fatal error from the include instead of checking if the file exists.

dlolsen:
I think it's a good idea to skip loading an addon with no `bootstrap.php`, but I totally get @aheinze comment about speed tradeoff.  Given this, I'm apt to close this pull request…

_**Unless**_ someone else knows how to properly catch the error in `bootModule` on line 1332 when `require($module->_bootfile);` fails to load a `bootstrap.php` file which doesn't exist.

I had tried to follow the ideas @Crease29 outlined in his comment and throw a `\RuntimeException`, but couldn't figure out how to catch a failed `require` statement. Since [`require`](http://php.net/manual/en/function.require.php) doesn't produce any return value on failure (rather it will _produce a fatal E_COMPILE_ERROR level error_) I can't do something like this:
```
try{
    if(!require($module->_bootfile)) throw new \RuntimeException('no bootstrap.php');
}
catch(\RuntimeException $e){
// quietly proceed
}
```
And, because the error is fatal, the script doesn't seem to make it to the throw statement. In the end it appears that the only way to catch this is to define a custom handler in either `set_error_handler` or `register_shutdown_function`, neither of which seems like the best idea, or something I feel ready to tackle.

If no one else has ideas, or it's not serious enough of an issue, I'll close the pull request and issue.

Crease29:
The only possible way to catch this error would be to have an own error handler in cockpit. That would need a whole new decision. But then you would be able to throw an Exception and catch it.

See an example here: https://stackoverflow.com/a/14103883

dlolsen:
I had considered a custom error handler as you mention, but if you define a custom error handler to catch potential non-existent `require` files, and even if you reset the error handler after checking, the custom error handler will still be used for the require files which actually do exists. This will likely cause unintended and unexpected results.

For this reason, and since I no one else has suggested alternate solutions. I'm closing the pull request.

Issue #1034, opened 2019-02-26
Upload of assets using the API doesn't accept meta


pauloamgomes:
When uploading assets using the API seems not possible to pass a meta information for the asset.

However, that is supported by the uploadAssets function:

```
'uploadAssets' => function($param = 'files', $meta = []) {
```

think that can be handy mostly when doing migration of contents (e.g. set assets to the proper folders, etc..)

PR #1035, opened 2019-02-26
#1034 - support for meta

pauloamgomes:
That should provide meta on api requests and therefore define meta fields (e.g. title, description, folder) during the asset creation using the api, e.g.:

image

### DIFF
--- a/modules/Cockpit/module/assets.php
+++ b/modules/Cockpit/module/assets.php
@@ -29,7 +29,7 @@ $this->module('cockpit')->extend([
         $assets    = [];
         $created   = time();
 
-        foreach ($files as &$file) {
+        foreach ($files as $idx => &$file) {
 
             // clean filename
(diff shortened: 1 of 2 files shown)


modules/Cockpit/module/assets.php
```diff
@@ -83,6 +83,10 @@
 
             $opts = ['mimetype' => $asset['mime']];
 
+            if (isset($meta[$idx]) && is_array($meta[$idx])) {
+                $meta = $meta[$idx];
```
aheinze: you're overriding the $meta array here. so it will only work for `$meta[0]` in the next iteration e.g. `$meta[1]` isn't available

pauloamgomes: ups, my bad :( 
should be fixed now

aheinze:
Thanks!

PR #1036, opened 2019-02-26
minor UI improvements

raffaelj:
see commits and commit descriptions


raffaelj:
I reverted my last commit to make the boolean field clickable again, because the switch itself wasn't clickable anymore. The whole problem appeared, since I changed `onclick` to `onchange` in #1007.

aheinze:
Thanks for the contribution. I'm not sure about the dashboard css yet. Need to investigate it further first.

raffaelj:
> I'm not sure about the dashboard css yet. Need to investigate it further first.

It looked like you wanted to revert the margin-top from `.app-main` with the negative margin of `dashboard-header-panel`. That's why I decided to keep it as it is, but only if `dashboard-header-panel` is the first child. As soon as some content is added above, it's not the first child anymore and the negative margin disappears.

Issue #1038, opened 2019-02-28
Enhancement on field repeater


pauloamgomes:
The field repeater has all items collapsed by default and that sound right, but only displays the label or type of the field and not the `display` value like in previewOrder mode, my suggestion is to apply also the display when items are collapsed.

So right now we can have something like:

<img width="450" alt="image" src="https://user-images.githubusercontent.com/102261/53565579-c0876800-3b51-11e9-9a96-88238666bbcf.png">

But would be better if we can have instead:

<img width="450" alt="image" src="https://user-images.githubusercontent.com/102261/53565543-abaad480-3b51-11e9-8dc8-d98256145acd.png">

so it would be:

`label/type - display field value`

Another addition would be to have also the text clickable in order to expand.

PR #1039, opened 2019-02-28
#1038 preview text on collapsed items

pauloamgomes:
Adds text from order preview to the collapsed repeater items


aheinze:
Thanks!

PR #1043, opened 2019-03-03
Fix Location component z-index

### LINKED ISSUE #1041: z-index issue on entry view - map overlays action bar
When you scroll down the page the map location field sits above the action bar, the z-index on the action bar needs to be increased.

cockpit-zindex-issue

Blistok:
Fixes #1041 

screen shot 2019-03-03 at 12 42 28 am


aheinze:
Thanks!

PR #1044, opened 2019-03-03
Tinymce fixes

raffaelj:
* added functionality to disable tinymce plugins - now pageurl, assetpath and mediapath are available as separate plugins --> Works also with https://github.com/pauloamgomes/CockpitCms-EditorFormats
* added i18n for tinymce - copy i18n file into `config/cockpit/i18n/tinymce/de.js`


aheinze:
Thanks!

Issue #1048, opened 2019-03-06
Provide a safe delete (with confirmation) on repeaters (configurable)


pauloamgomes:
Repeaters are a nice feature for aggregating contents when using with simple elements (e.g. textfield) a delete is more or less harmful, but when using with many nested fields (set) we can remove by mistake and there is no way to recover the deleted item. My proposal is to have a confirmation popup before the delete operation so it will avoid accidental removals. Such an option should be configurable in the field definition so it can be used on only relevant fields.

PR #1049, opened 2019-03-06
update for #1048 - safe delete

pauloamgomes:
#1048 provides  a confirmation on the delete operation when we have a safeDelete attribute


aheinze:
Thanks!

PR #1064, opened 2019-03-21
Allows getting the value at path of object in field-collectionlink

DrMartinGonzo:
This enables us to get a proper name when linking to a collection having a set-like field or an asset field for example.
This replicates the behaviour seen in field-repeater.
What do you think ?


aheinze:
Thanks!

Issue #1062, opened 2019-03-19
Suggested Enhancement - "Range" or "Number" field type


lmlsna:
It seems like there are only 2 field types that deal with numeric input, regular old text inputs _(which works just fine with a `([0-9]*)` pattern restriction)_ or the `rating` element _(which rapidly becomes absurd with maximum values over 10 as it displays a star for each number)_.

There are 2 HTML5 native elements that would be more functional and could be implemented with very small modifications to existing field types:

- A [`range`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range) input _(which renders out of the box as an ugly slider on most browsers)_ could be created by cloning the [field-rating](https://github.com/agentejo/cockpit/blob/next/modules/Cockpit/assets/components/field-rating.tag) field and making a few changes (mostly removing stuff).

- A [`number`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) input field type (a textbox for fallback with a little up/down button next to it) could be implement with a few changes to the standard [field-text](https://github.com/agentejo/cockpit/blob/next/modules/Cockpit/assets/components/field-text.tag) input.

I'm not a big JS guy, otherwise I would have just sent a pull request straight away. I'm sure I _could_ manage a medium shitty pull request if you think this is a useful idea, and no one else feels like doing it.

But I figured someone else who knows the codebase could probably do that about 50x faster than me. Also just wanted to make sure I wasn't missing anything and that this would be useful before spending the time... It's a lot easier to use this than to hack on it lol.

Any thoughts?

Blistok:
@lmlsna The Number field type:
https://github.com/Blistok/CockpitCMS-NumberField

aheinze:
a workaround would be to use a text field with `"type":"range"` in the field options

Issue #822, opened 2018-07-20
[request] additional field types


RoelantStegmann:
Is it possible to extend the field types with some additional ones?

Integer / Numeric
Use case: we want to keep some prices in the database, but want to enforce it to be numeric. It are standard JSON datatypes so it feels like they would fit. 

RepeaterSet
Use case: we want to for example add multiple alt- title - image combinations, or firstname - lastname combinations to one object. Respectively images for a product or stores with two contact persons.  For this you would want to use Repeater on a combination of several fields.

DigitalGoldfish:
Regarding the RepeaterSet - this is something that you can already do by putting a set field inside a repeater. You will have to do the configuration in the options field ... there is no UI for this.

Here's an example:
`  "field": {
    "type": "set",
    "name": "links",
    "label": "Links",
    "options": {
		{
		  "fields": [
			{
			  "name": "name",
			  "label": "Name",
			  "type": "text",
			  "localize": true
			},
			{
			  "name": "url",
			  "label": "Adresse",
			  "type": "text",
			  "localize": true
			},
			{
			  "name": "type",
			  "type": "select",
			  "label": "Link-Typ",
			  "options": {
				"options": "Webseite, Email",
				"default": "Webseite"
			  }
			}
		  ]
		}
	},
	"display": null
}`

P.S.: I usually just edit the PHP schema files directly bypassing the UI.

piotr-cz:
This is how I've build field to create multiple files with label based on @DigitalGoldfish answer
```json
{
  "field": {
    "type": "set",
    "name": "items",
    "label": "Item",
    "options": {
      "fields": [
        {
          "name": "label",
          "label": "Label",
          "type": "text"
        },
        {
          "name": "file",
          "label": "File",
          "type": "file"
        }
      ]
    }
  }
}
```

ronbravo:
@RoelantStegmann Not sure if this helps but, I've been updating the schema by API Rest call. I just added a type like so: `{ name: 'total', type: 'number', info: 'The display name for the user.' }` Then when I inspect the entry in the SQLLite DB it seems to be saving the entry for total as a number.

For example:
```
{
   "alias":"toni taylor",
   "email":"toni.taylor@example.com",
   "image":"https:\/\/randomuser.me\/api\/portraits\/women\/55.jpg",
   "username":"goldenladybug322",
   "total":667,
   "_by":null,
   "_modified":1545413933,
   "_uid":"",
   "_creatorId":"",
[...]

PR #1068, opened 2019-03-23
[WIP] Add number field

Blistok:
WIP - Please don't merge @aheinze.

This PR adds explicit `number` field. I'm aware of the option to use `text` field for this purpose (by adding `"type": "number"` in options (related discussions: [link 1](https://discourse.getcockpit.com/t/number-field-type/393/2), [link 2](https://github.com/agentejo/cockpit/issues/822#issuecomment-458880616)).
The issue I'm facing with current solution is that, even though the UI recognizes the type number field and renders appropriate input tag, the actual value gets converted in 2 places from number to string.
First (implicit conversion) happens when doing JSON serialization [here](https://github.com/agentejo/cockpit/blob/next/modules/Collections/views/entry.php#L11), and the second (explicit) one just before saving [here](https://github.com/agentejo/cockpit/blob/next/modules/Collections/bootstrap.php#L302).
Also, it seems that @aheinze  already had this use case in mind judging by [this commit](https://github.com/agentejo/cockpit/commit/4a9057541de1b557bdc92e9ea1497a3609732d16) and [this comment](https://discourse.getcockpit.com/t/api-returns-int-for-text-value/366/8).

So it seems that we have 2 options, either to fix this return type or to switch to having explicit `number` field type (this PR). Pro of the first approach is that it would be less code (probably could be done with one line surgically in particular place). Pros of the second approach are, even though we end up with more code, we have benefit of transparency (the whole discussion from link 1 and link 2 above wouldn't be happening in the future) and by not casting values (especially in multiple places) we would avoid source of subtle bugs (it took me a while to figure out why my sort function on one of the "number" fields didn't give expected results on the frontend before I realize that it wasn't sorting by number values but by ASCII string values!).
Finally, should we decide to go this route we'll have to consider plugins support. For example, CockpitQL would have to update it's [types list](https://github.com/agentejo/CockpitQL/blob/master/Types/FieldType.php#L37-L49) before it could work with the new number field.
 
Related issue: #1062

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -3212,6 +3212,21 @@ riot.tag2('field-multipleselect', '<div class="{options.length > 10 ? \'uk-scrol
 
 });
 
+riot.tag2('field-number', '<div class="uk-position-relative field-number-container"> <input ref="input" class="uk-width-1-1" bind="{opts.bind}" placeholder="{opts.placeholder}" type="number"> </div>', '', '', function(opts) {
+        this.on('mount', function() {
+
+            opts.cls && App.$(this.refs.input).addClass(opts.cls);
+
+            opts.required && this.refs.input.setAttribute('required', 'required');
+
+            ['max', 'min', 'placeholder', 'readonly', 'step'].forEach( function(key) {
+                opts[key] && this.refs.input.setAttribute(key, opts[key]);
+            }.bind(this));
+
+            this.update();
+        });
+});
+
 riot.tag2('field-object', '<div ref="input" riot-style="height: {opts.height || \'300px\'}"></div>', '', '', function(opts) {
 
         var $this = this, editor;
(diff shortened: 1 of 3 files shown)


aheinze:
is this still needed after this commit: https://github.com/agentejo/cockpit/commit/781c44e813d64dfecc3ffb5eab372d75937d7cfb ?

Blistok:
So, that one seems to solve the implicit conversion (the first one I mentioned in the original description), but then when it gets saved it gets explicitly converted from number to string [here](https://github.com/agentejo/cockpit/blob/next/modules/Collections/bootstrap.php#L302).

Simple test case is to put one console.log for `this.entry` just before [this line](https://github.com/agentejo/cockpit/blob/next/modules/Collections/views/entry.php#L313) and one for `entry` inside .then() when promise resolves after saving the entry; and you should see in the first log it will return number, and in the second it will be converted to string.

Blistok:
@aheinze I'm gonna close this PR. I ended up doing separate mini addon for the number field here:
https://github.com/Blistok/CockpitCMS-NumberField

PR #1071, opened 2019-04-02
Improve custom form validation response message

willbrowningme:
Currently when adding custom form validation as described in the docs you can only return false if a certain field fails validation.

```php
if ($data["field1"] != "value1") {
    return false;
}

return true;
```

This pull request would allow you to return a custom validation error message like so:

```php
if ($data["field1"] != "value1") {
    return "Field1 did not have the correct value, please try again.";
}

return true;
```

This allows you to display this error message on the client side and give a user more information on why the form failed validation.

I added the conditional `return false===$validation ? false : ['error' => $validation, 'data' => $data];` so that this change will not be breaking for anyone still using the current validation response.

### DIFF
--- a/modules/Forms/bootstrap.php
+++ b/modules/Forms/bootstrap.php
@@ -260,8 +260,8 @@ $this->module('forms')->extend([
         }
 
         // custom form validation
-        if ($this->app->path("#config:forms/{$form}.php") && false===include($this->app->path("#config:forms/{$form}.php"))) {
-            return false;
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks for the contribution, but I don't think this is needed. What you can also do in the validation file:

```
$this->app->stop(['error' => 'Field1 did not have the correct value!'], 412);
```

`412` is the http error code for 'Precondition Failed'

willbrowningme:
You're right that is cleaner. I'm still learning the ins and outs of Cockpit!

PR #1076, opened 2019-04-05
1075 - implements login error trigger

pauloamgomes:
Adds a new trigger for errors during login passing only the username.


aheinze:
Thanks!

PR #1077, opened 2019-04-06
Fix #1074

### LINKED ISSUE #1074: [security] Filters are not being applied when pointing directly to the admin entry URL
Hi, I realized that if I go directly to the entry URL from the admin panel my filters (collection.read.php) don't take effect, they work on the lists, counters, etc. but not when going directly to the URL, Am I missing something?

I created a custom group and enabled just the View Entries permission. I'm using MongoDB.

E.g. URL: /collections/entry/articles/5ca76d2ea4c0a53a443cbd20

I guess that this happens because **Collections\Controller\Admin->entry()** doesn't do any filtering?

I realized that collection.read.php gets executed but 'too late'.

Thank you.

filisko:
This would fix #1074

### DIFF
--- a/modules/Collections/Controller/Admin.php
+++ b/modules/Collections/Controller/Admin.php
@@ -208,11 +208,11 @@ class Admin extends \Cockpit\AuthController {
 
         if ($id) {
 
-            //$entry = $this->module('collections')->findOne($collection['name'], ['_id' => $id]);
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks for the contribution. Need to check first if there are some side-effects

filisko:
Yes, side-effects are the main concern. 👍

filisko:
No updates on this? :thinking:

aheinze:
Thanks again!

Issue #755, opened 2018-05-07
Required fields option not working


lilybarrett:
I've attempted to designate some form fields as required by sliding the interface's "Required" button for said fields over to the right, turning the associated buttons green. The desired behavior here is that an error would be produced if the user attempts to save an entry without filling out required fields. However, the form saves successfully even when those fields are left untouched.

We'd like a reliable way to ensure that the user is required to fill out certain fields.

aheinze:
Hi,

which field types did you use?

lilybarrett:
The fields in question are both Text types. Out of curiosity, why should that matter?

gryphonmyers:
I am also experiencing a failure to enforce requirements, but my use case is a little more fringe so may not be currently supported - I'm attempting to add required flag to a nested set field.

stefanKBL:
hi,
similar issue when post a new entry via front-end or postman, using collections/save/collectionname
just setting a username field as required, sending data with blank username, and no error, entry is created in collection... without username!
Solution is of course to do validation in front-end, but i think it must also be done by cockpit, no?
Thanks

PR #1080, opened 2019-04-10
Fixes collection required fields

sa7bi:
This pull request should fix issue #755 where empty required fields are saved to the DB.


aheinze:
Thanks! But I see some edge cases for the current check. e.g. what if it is a numeric field and the `$value` is `0`. Then it would also invalidate even if it is a correct/possible value. maybe check with `empty`? 🤔

Issue #1089, opened 2019-04-24
feature request - make repeater items clonable


pauloamgomes:
In some situations, when using a repeater with set, having the possibility to clone existing items is quite useful, e.g. we have n fields and only 1 or 2 will be differ, so we can clone a previous item).

pauloamgomes:
Example of desired functionality:
https://monosnap.com/direct/uOEz1faqJnquHGOZyl9NpaaWUiWr87

PR #1090, opened 2019-04-24
support for cloning

pauloamgomes:
addresses #1089


aheinze:
Thanks for the contribution! I would remove the option `clonable`, this isn't really needed

pauloamgomes:
> Thanks for the contribution! I would remove the option `clonable`, this isn't really needed

You mean have the clone option by default?

aheinze:
Thanks!

PR #1095, opened 2019-05-02
Improved UX while editing in preview mode

ebspeter:
This is shifting the modal to left side and removing the backdrop in preview mode. 

Example:
image


aheinze:
I'm not sure about that. for now you can add this via a custom addon

ebspeter:
This tiny style change improves the editing experience significantly IMHO, it was the first request i got from our content team when i introduced Cockpit, so I imagine others might benefit too. 

Is there any benefit of having the modal on top of the preview of the content you are currently editing?

A few lines of css doesn't feel like enough meat for a custom addon, maybe keep this open for a while to see if anyone else want this. Then they'll also have a simple copy/paste if they want to implement in their own fork.

fredrikkarlsson-pervanovo:
I also wanted this, so I created a addon for it: https://github.com/Pervanovo/cockpit-BetterPreviewModal

PR #1097, opened 2019-05-07
boolean field rewrite

raffaelj:
* semantic markup - label is now an actual, clickable label
* removed some js - not necessary anymore
* changed CSS
  * moved switch into :before/:after
  * minor changes (em instead of px)
  * added :focus for visibility when using tab key


raffaelj:
Clicking on the boolean field label is possible again after I changed the switch from `onclick` to `onchange` in #1007.

I addressed this before in #1036, but reverted it later. Now I changed the CSS properly.

aheinze:
Thanks!

Issue #1100, opened 2019-05-08
Improve locking by providing the ability (per permission) to break locks


pauloamgomes:
Based on the discussion on discourse (https://discourse.getcockpit.com/t/collaborative-work-user-tracking-how-to-prevent-editing-the-same-entry-simultaneously/626/14).

The idea is to provide the ability via global permission to break existing locks.

PR #1101, opened 2019-05-08
simple unlock ui based on global permission

pauloamgomes:
Addresses idea from #1100 using a simple approach, not sure if it would make sense to have more granular options (permission for accounts, singletons, collections), but that would only increase unnecessary complexity.


pauloamgomes:
Other thing is right now an superadmin unlocks a content just by accessing it, I think would make more sense to have the unlock button available by default.. so its up to the admin if he wants to take over on the content or not.

aheinze:
Thanks for the contribution! I think this is a good starting point. I'll to add some improvements.

PR #1105, opened 2019-05-10
Add installation check for /storage/thumbs.

danielbecker:
I added a check for the /storage/thumbs folder as it is necessary for a working installation.


aheinze:
Thanks!

PR #1108, opened 2019-05-10
During installation, create missing folders when possible.

itbrandonsilva:
Awhile ago, I encountered an issue when attempting to use cockpit in a containerized environment. If some folder(s) did not exist before initiating installation, the install would be unable to proceed. This update allowed us to run through the cockpit install despite mounting a totally empty volume.


aheinze:
Thanks for the contribution!

PR #1128, opened 2019-06-04
Cleanup

raffaelj:
I fixed some minor things, that disturbed me ;-)

* rounded borders for single buttons in button group when creating collection/singleton
* a single copy-paste typo "clobal" everywhere - that typo still exists in multiple addons


aheinze:
Thanks for the contribution, can you please move the button styling to `/assets/app/css/less/uikit/button.less` - https://github.com/agentejo/cockpit/blob/next/assets/app/css/less/uikit/button.less

raffaelj:
I moved it.

One question: Are the files in `/assets/app/css/less/uikit` from Uikit or are they handwritten? I'm always afraid of changing anything there, because it might get overwritten when updating uikit.

aheinze:
the uikit core files are in `/assets/lib/uikit`, the app overrides for uikit are in `/assets/app/css/less/uikit/`

aheinze:
Thanks!

PR #1129, opened 2019-06-04
cleanup

raffaelj:
I knew, there was a completely "broken" class somewhere and I stumbled upon it again. I replaced all tabs with spaces and fixed the indentation here and there. A few functions were unreadable, because my editor displays tabs like 2 spaces.


aheinze:
Thanks!

Issue #1135, opened 2019-06-20
Enhance filter options on the entries ui


pauloamgomes:
I'm aware that can use the events `collections.find.before` and `collections.find.before.$name`, but I want just to provide custom filtering on the collections page ui.

My suggestion is to trigger an event that changes the options in the find() method of the Admin controller.

PR #1137, opened 2019-06-24
added event for manipulate admin find options

pauloamgomes:
#1135  Having the event here is handy as can provide to the users further search in other fields than the title.


aheinze:
could you describe a use-case?

pauloamgomes:
I have a collection that I need editors to search by referenced contents.
Or search by slug.

aheinze:
could you then also please add an `collections.admin.find.after.{$collection['name']}` event. Thanks.

aheinze:
Thanks!

PR #1139, opened 2019-06-27
Asset upoad from CloudStorage plugin cannot close stream

fredrikkarlsson-pervanovo:
When uploading an asset to google cloud storage, the stream cannot be closed.
The simplest solution is to only close stream if it is a resource.

This is needed to make https://github.com/agentejo/CloudStorage/pull/6 not give frontend error messages when uploading an asset.


aheinze:
Thanks!

fredrikkarlsson-pervanovo:
Thanks for a very promising cms platform ;)

PR #1143, opened 2019-07-02
fixed wrong padding if text field is not type=text

raffaelj:
If a text input field has a different type than "text", it had a hard coded padding for the length indicator. Now the padding is only applied, if it is a text field.

### DIFF
--- a/modules/Cockpit/assets/components/field-text.tag
+++ b/modules/Cockpit/assets/components/field-text.tag
@@ -2,7 +2,7 @@
 
     <style>
 
-        [type=text] {
+        [ref="input"][type=text] {
             padding-right: 30px !important;
         }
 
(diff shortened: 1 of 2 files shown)


modules/Cockpit/assets/components/field-text.tag
```diff
@@ -2,7 +2,7 @@
 
     <style>
 
-        [ref="input"] {
+        [type=text] {
```
aheinze: could you please leave the `[ref="input"]` here. so the change should look like this:

`[ref="input"][type="text"]`

raffaelj: OK, I changed it.

aheinze:
thanks!

PR #1159, opened 2019-08-07
Fix Call to undefined function levenshtein_utf8()

piotr-cz:
When using $fuzzy/ $text filters, error comes up:
```
Error: Call to undefined function levenshtein_utf8()
```

This is because _levenshtein_utf8_ is not a global function and should not be called via `\levenshtein_utf8()`

Bug was introduced in 411ec6e8e6


aheinze:
Thanks!

PR #1160, opened 2019-08-07
Fix php 7.2 deprecation notice in $mod filter

piotr-cz:
When using mod filter, deprecation message comes up:
```
The each() function is deprecated. This message will be suppressed on further calls
```
This is because `each` has been [deprecated in php 7.2](https://www.php.net/each)


aheinze:
Thanks!

PR #1162, opened 2019-08-07
use utf8 expression for case insensitive regex search, fixes #1161

### LINKED ISSUE #1161: MongoLite - regex search is not case insensitive for special chars
Steps to reproduce:

* Create a collection entry with title "Über uns"
* Go to entries view via admin ui and search for "über" --> no results
* Search for "Über" --> 1 result

raffaelj:


aheinze:
Thanks!

PR #1164, opened 2019-08-08
Fix $mod arguments

piotr-cz:
At this moment $mod arguments signature in MongoLite driver is
```
['$mod' => [<divisor> => <ignored>]]
```

As per [MongoDB docs](https://docs.mongodb.com/manual/reference/operator/query/mod/) it should be 
```
['$mod' => [<divisor>, <remainder>]]
```

while remainder is required, I've used fall back to 0


aheinze:
Thanks!

PR #1166, opened 2019-08-08
Fix $not operator (MongoLite)

piotr-cz:
As per [MongoDb Docs](https://docs.mongodb.com/manual/reference/operator/query/not/) `$not` operator isn't the same as `$neq`.

I've added the Regular expression functionality via `$regexp`

#### via Regular expression
```php
$app->storage->findOne('collections/test', 
    // When using MongoHybrid\MongoLite driver
    'content' => ['$not' => 'Lorem .*'],
    // When using MongoHybrid\Client driver
    // 'content' => ['$not' => new \MongoDB\BSON\Regex('Lorem .*')],
]);
```

#### via Operator (not implemented, for reference)
```php
$app->storage->findOne('collections/test', 
    'content' => ['$not' => ['$eq' => 'Lorem ipsum']],
]);
```

### DIFF
--- a/lib/MongoLite/Database.php
+++ b/lib/MongoLite/Database.php
@@ -394,7 +394,11 @@ class UtilArrayQuery {
             case '$regex' :
             case '$preg' :
             case '$match' :
+            case '$not':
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks!

PR #1172, opened 2019-08-20
make filter work on collection-link field

jonas8:


aheinze:
Thank you for your contribution, but can you split please your commits in separate pull requests?
I would merge some of your changes right away, but need to think about some others.

Thanks!

PR #1173, opened 2019-08-21
Fix SQLite DSN when using :memory: as a path

piotr-cz:
At the moment when using database server setting `mongolite://:memory:`, MongoLite Client creates new Database in format `:memory:/collectionName.sqlite` which throws PDOException.

I've changed the behavior to not append collection name when path is _:memory:_.
Created databases are still available in cache (_Client::$databases_) and it's possible to list them via _Client::listDBs()_ method

Fixing the _:memory:_ path this helps with creating tests for MongoLite driver.

### DIFF
--- a/lib/MongoLite/Client.php
+++ b/lib/MongoLite/Client.php
@@ -33,7 +33,7 @@ class Client {
     /**
      * Constructor
      *
-     * @param string $path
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks!

PR #1177, opened 2019-09-07
Missing settings event in forms module

raffaelj:
added forms.settings.aside event to enable the in_menu feature


aheinze:
Thanks!

PR #1179, opened 2019-09-11
Use overriden default language in "Copy content from"

fredrikkarlsson-pervanovo:
The default language name in config is not used as label for "Copy content from"


aheinze:
Thanks!

PR #1185, opened 2019-10-08
Add configurable path to compiled assets

piotr-cz:
At this moment the `#tmp` path is used for
- compiled views (`*.lexy.php`)
- webhooks cache (`webhooks.cache.php`)
- compiled assests (`assets/*.js`)

While first two items may be used in publicly inaccessible path on server, compiled asset files must be publicly accesible.

Usually I configure cockpit to store sensitive data in folders outside of the public directory (`#storage`, `#data`, `#cache`) and rest in public subdirectory (`#thumbs`, `#uploads`).

The problem with the `#tmp` path is that it mixes both (public and non-public files) so this PR adds new path `#compiledAssets`.


piotr-cz:
Please don't merge yet, there is some problem with file system resolving paths

piotr-cz:
Okay, it was a typo in in path on my side, fixed in https://github.com/agentejo/cockpit/pull/1185/commits/2bf014e4597fc9f4ae7068bad937a0af3de5f71c

aheinze:
I think `compiledAssets ` is to specific, I would rather prefer eg `publictmp` or do you have a better suggestion?

piotr-cz:
Let's think about about it from the other side. How would you call output directory?

piotr-cz:
Thanks

PR #1091, opened 2019-04-25
Expose parent options to nested Layouts/LayoutComponents

ebspeter:
Exposing settings being added in agentejo/LayoutComponents#4.

PR #1188, opened 2019-10-11
Merge parent component options with field-layout opts

fredrikkarlsson-pervanovo:
This is related to options in custom layout components (https://github.com/agentejo/LayoutComponents/pull/4)
and is another take on https://github.com/agentejo/cockpit/pull/1091 but without adding the options to the item (instance) data.

What this does is to merge the options in the parent component into the field-layout that is rendered to contain children. This allows to create addons to use the 'field.layout.components' event together with options created on a custom layout component.

The https://github.com/pauloamgomes/Cockpit-CustomComponents addon already uses this event and works as expected on the root field-layout but does not work on components with children since they do not receive opts from from either the layout definition or from component options.


aheinze:
Thank you for the contribution and sorry for the late answer! I'll test it as soon as possible.

modules/Cockpit/assets/components/field-layout.tag
```diff
@@ -229,6 +229,9 @@
             this.components = App.$.extend(true, this.components, window.CP_LAYOUT_COMPONENTS);
         }
 
+        if(opts.parentComponent && opts.parentComponent.options) {
+            opts = App.$.extend(true, opts, opts.parentComponent.options);
```
aheinze: shouldn't  it be the other way around  `opts = App.$.extend(true, opts.parentComponent.options, opts);`? that way the component options  can override the parent component options.

fredrikkarlsson-pervanovo: Yes it should, well spotted!

aheinze:
Thanks!

PR #1192, opened 2019-10-21
Localize collection entries page

fredrikkarlsson-pervanovo:
The collection entries page is lacking support for localized language.
This PR adds a language selector (like the one on the entry page) and adds the lang option to collection loading.

Also included is filtering with localized entries. It creates a filter object (instead of string) and creates $or criteria (pretty much like the backend does when it receives a string) but adds the language suffixes for localized fields.

### DIFF
--- a/modules/Collections/views/partials/entries.php
+++ b/modules/Collections/views/partials/entries.php
@@ -114,6 +114,19 @@
             </div>
         </div>
 
+        <div class="uk-margin-top uk-text-center uk-form" if="{ !loading && languages.length }">
+            <div class="uk-form-select">
+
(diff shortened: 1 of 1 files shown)


aheinze:
Thank you for the contribution, I'll test it as soon as possible.

aheinze:
I'll close this PR because of a slightly different implementation in https://github.com/agentejo/cockpit/commit/b66c66eeb28527f71905468038970328340e9751 . 
But thank you for pushing the idea 👍

fredrikkarlsson-pervanovo:
Great! I avoided to change the backend on purpose, but this seems like a much more elegant solution.

PR #1196, opened 2019-10-29
Collection live preview i18n lang fix

fredrikkarlsson-pervanovo:
Collection live preview should:
* keep the selected language from the collection entry page
* display the default language label instead of the i18n

This PR fixes both issues


aheinze:
Thanks!

PR #1197, opened 2019-10-30
Bugfix: Add new layout component in correct place

fredrikkarlsson-pervanovo:
In collection entry with a layout field:
When pressing the centered "Add" button (plus symbol), the new item gets placed using the parent elements index, which is not correct since it should be added last in the context.
Pressing the hidden add symbol on an element works as expected.


aheinze:
Thanks!

fredrikkarlsson-pervanovo:
@aheinze Wow! You are quick today :D

PR #1203, opened 2019-11-13
add href attribute to app search

### LINKED ISSUE #1202: allow middle click for app search results to open entry in new tab
I like browsing with multiple open tabs. When I use the app search to find collections etc., I have to click with the left mouse button and I can't use middle click to open it in a new tab.

raffaelj:
fixes #1202


aheinze:
Thanks!

Issue #1204, opened 2019-11-13
how to close app-panel-dropdown isn't obvious


raffaelj:
Steps to reproduce:
* click on the site logo at the top left --> the system menu appears

Users have to guess, that they have to click in the empty area of the app panel dropdown to close it and

1. It is impossible to close the dropdown without a mouse (accessibility).
2. Closing fails if a user clicks in the empty area of a `<ul>`
3. Even if it is not necessary, a close icon would be a visible helper.
4. I would like to close it via Escape key, but that's not so important.

aheinze:
Commit fixed 1,2 + 4 ... I don't like the close icon idea. I think implementing close on esc is also intuitive

raffaelj:
OK. Thanks for the fix.

PR #1205, opened 2019-11-13
added close icon, always close panel on click

raffaelj:
fixes point 1-3 of #1204


aheinze:
Thanks for the contribution Raffael. But I have to close this PR because of https://github.com/agentejo/cockpit/issues/1204#issuecomment-554111863

PR #1209, opened 2019-11-25
Add value attribute to text fields

sergeysedoy97:


aheinze:
What is the reason behind this PR?

sergeysedoy97:
Sometimes we need default value for text fields.
Attribute condition fails if some of attributes is 0 (number).

aheinze:
why not don't you use the `default` field setting?

PR #1216, opened 2019-12-15
Added $file to cockpit.asset.upload trigger

fredrikkarlsson-pervanovo:
I am creating an addon to modify assets before they are uploaded and I need to get the path to the tmp file. Using the data available in the cockpit.asset.upload event, I cannot find the tmp files without passing the $file variable.


aheinze:
Thanks!

PR #1219, opened 2019-12-23
Encapsulate app bootstrap

raulferras:
The main point of this PR is being able to integrate Cockpit more easily with an existing infrastructure. Specifically:
- Be able to place the code outside public folder (even on `vendor`).
- Place the entry point (index.php) outside of the project folder.

This has been achieved by creating `\Cockpit\App` which is responsible to boot and run the cockpit application in a more encapsulated way.

`\Cockpit\App` has a more explicit contract and runs away from the "extensiveness" `bootstrap.php` allowed, which contradictorily makes customisation difficult (for example setting paths required both setting constants + configuration values) and not very predictable.

Basically `\Cockpit\App` contains the logic of `bootstrap.php` (a little modified to simplify some points), but it also contains logic that was present in `index.php` and `cp.php` that I think belongs to this new domain.  
This even allows installing cockpit with a plain `composer require` and run it from the `vendor` folder when installed with composer.

There are some deprecated behaviours in this PR:
- bootstrap.php file has been removed.
- COCKPIT_* constants are delegated to be only defined by Cockpit itself. As their values are infered from configuration values/$_SERVER globals, there is no reason to allow developer to pre-define them.
- Cockpit is no longer responsible to read the config.php/config.yml files.
   The configuration array is just injected in the App constructor. This gives freedom to use whatever method user wants to use to define their configuration.

Other changes:
- Registered PSR-0 folders in composer.json. This allows to get rid of some hardcoded `require` calls.
- Document some COCKPIT_* constants.
- Simplify some of the initialisation variables (use more meaningful names and avoid creation of temporal variables that are quickly discarded).


piotr-cz:
This is very interesting, good job!
This would close few issues about composer installation

By observations:

- Removing `bootstrap.php` file is a backward compatibility break (for example in my projects I'm requiring cockpit in the frontend app via this file).
  I would keep it, but add deprecation comment at the top, Otherwise change this should come in v2.0.

- I don't like the idea of symlinking assets, but this may be personal.
  Maybe there could be new cli command which would copy them to specified directory?

raulferras:
> Removing bootstrap.php file is a backward compatibility break (for example in my projects I'm requiring cockpit in the frontend app via this file).
I would keep it, but add deprecation comment at the top, Otherwise change this should come in v2.0.
I agree.

> I don't like the idea of symlinking assets, but this may be personal.
Maybe there could be new cli command which would copy them to specified directory?
I can check creating a command that both copies or creates symlinks to it.

PR #1243, opened 2020-02-09
fix link to '/accounts' for users without access, fixes #1214

### LINKED ISSUE #1214: 401 when "Cancel" on SETTINGS > ACCOUNTS > ACCOUNT
The "Cancel" button/link at the bottom of the page at `/accounts/account` is currently hardcoded to go to the `/accounts` page regardless of the whether or not the current user has permission to view the `/accounts` page. For some users, a 401 Not Authorised message is shown when they press "Cancel", which feels like a confusing experience.

This could be avoided by checking to see if the user has permission to view the `/accounts` page a bit earlier. I suggest changing [the template that generates the  "Cancel" button/link](https://github.com/agentejo/cockpit/blob/next/modules/Cockpit/views/accounts/account.php#L115) from this:

```
<a class="uk-button uk-button-large uk-button-link" href="@route('/accounts')">@lang('Cancel')</a>
```

to this:

```
<a class="uk-button uk-button-large uk-button-link"
@hasaccess?('cockpit', 'accounts') href="@route('/accounts')" @else href="@route('/')" @endif
>@lang('Cancel')</a>
```

raffaelj:


aheinze:
Thanks!

Issue #1244, opened 2020-02-11
rendered Lexy templates mess up indentation


raffaelj:
I like the Lexy template parser, but I don't like the messed up html indentation.

**fix:**
https://github.com/raffaelj/cockpit/commit/9ad961ebe5ca4b3472ffc396d5fde93f518f29d7

Find php tags after line breaks with following white space and replace them without the white space. Multiple line breaks are converted to a single linebreak. The regex fails for php tags with white space at the first line of the file.

```php
preg_replace('/([\r\n])(\s*)\<\?php/', '$1<?php', $value);
```

**example template:**

```php
    <p>one line test @if(1==1) one @elseif(2==2) two @else three @endif</p>
    <p>nested multi line test:</p>
    {% $arr = [1,2,3]; %}
    <ul>
    @if(count($arr))

      @foreach($arr as $i)
        <li>
        @if($i==1)
        one
        @elseif($i==2)
        two
        @else
        three
            <ul>
              @foreach([1,2] as $v)
                <li>{{ $i*$v }}</li>
              @endforeach
            </ul>
        @endif
        </li>
      @endforeach

    @endif
    </ul>
```

## current output

**rendered php:**

```php
    <p>one line test <?php if (1==1) { ?> one <?php } elseif(2==2) { ?> two <?php } else { ?> three <?php } ?></p>
    <p>nested multi line test:</p>
    <?php $arr = [1,2,3]; ?>
    <ul>
    <?php if (count($arr)) { ?>

      <?php foreach ($arr as $i) { ?>
        <li>
        <?php if ($i==1) { ?>
        one
        <?php } elseif($i==2) { ?>
        two
        <?php } else { ?>
        three
            <ul>
              <?php foreach ([1,2] as $v) { ?>
                <li><?php echo  $i*$v ; ?></li>
              <?php } ?>
            </ul>
        <?php } ?>
        </li>
      <?php } ?>

    <?php } ?>
    </ul>
```

**ugly html output:**

```html
    <p>one line test  one </p>
    <p>nested multi line test:</p>
        <ul>
    
              <li>
                one
                </li>
              <li>
                two
                </li>
              <li>
                three
            <ul>
                              <li>3</li>
                              <li>6</li>
                          </ul>
                </li>
      
        </ul>
```

## output after my fix

**rendered php:**

```php
    <p>one line test <?php if (1==1) { ?> one <?php } elseif(2==2) { ?> two <?php } else { ?> three <?php } ?></p>
[...]

PR #1245, opened 2020-02-11
trim whitespace between php tags after line breaks

raffaelj:
see #1244

### DIFF
--- a/lib/Lexy.php
+++ b/lib/Lexy.php
@@ -28,7 +28,8 @@ class Lexy {
         'else',
         'unless',
         'unescape_echos',
-        'php_tags'
+        'php_tags',
+        'cleanup'
     );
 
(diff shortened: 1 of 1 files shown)


aheinze:
Thank you for the contribution Raffael! 
But because this is more an opinionated cosmetical improvement, I would like to politely decline this PR.

You can still achieve this btw by adding the following snippet to an addon: 

```
cockpit()->renderer->extend(function($content) {
    return preg_replace('/([\r\n])(\s*)\<\?php/', '$1<?php', $content);
});
```

raffaelj:
I understand, that you decline it. The problem is, that I can't extend it that way. The user extensions are called before the internal compile functions, so it won't have any effect. I tried this in the first place before changing Lexy itself.

PR #1246, opened 2020-02-13
Added option to set collection entries sorting

piotr-cz:
This adds an option to set collection entries sorting in the collection screen.
You may choose from one of the fields and direction.

image

### DIFF
--- a/modules/Cockpit/assets/components/field-select.tag
+++ b/modules/Cockpit/assets/components/field-select.tag
@@ -2,7 +2,7 @@
 
     <select ref="input" class="uk-width-1-1 {opts.cls}" bind="{ opts.bind }">
         <option value=""></option>
-        <option each="{ option,idx in options }" value="{ option.value }" selected="{ parent.root.$value === option.value }">{ option.label }</option>
+        <option each="{ option,idx in options }" value="{ option.value }" selected="{ parent.root.$value == option.value }">{ option.label }</option>
     </select>
 
     <script>
--- a/modules/Collections/Controller/Admin.php
+++ b/modules/Collections/Controller/Admin.php
@@ -76,6 +76,8 @@ class Admin extends \Cockpit\AuthController {
             'fields'=>[],
             'acl' => new \ArrayObject,
             'sortable' => false,
+            'sort_column_name' => '_created',
+            'sort_direction' => -1,
             'in_menu' => false
         ];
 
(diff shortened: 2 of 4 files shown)


modules/Cockpit/assets/components/field-select.tag
```diff
@@ -2,7 +2,7 @@
 
     <select ref="input" class="uk-width-1-1 {opts.cls}" bind="{ opts.bind }">
         <option value=""></option>
-        <option each="{ option,idx in options }" value="{ option.value }" selected="{ parent.root.$value === option.value }">{ option.label }</option>
+        <option each="{ option,idx in options }" value="{ option.value }" selected="{ parent.root.$value == option.value }">{ option.label }</option>
```
aheinze: what is the reason for this change?

modules/Collections/Controller/Admin.php
```diff
@@ -76,6 +76,8 @@ public function collection($name = null) {
             'fields'=>[],
             'acl' => new \ArrayObject,
             'sortable' => false,
+            'sort_column_name' => '_created',
+            'sort_direction' => -1,
```
aheinze: can we have on config property here? 

```
sort: {
  column: "_created",
  dir: -1
}

modules/Cockpit/assets/components/field-select.tag
piotr-cz: This is required as field-select values are casted to strings , while sort direction is a number.
With strict comparison sort direction would have to be string

modules/Collections/Controller/Admin.php
piotr-cz: Probably yes, I'll check

piotr-cz: Resolved in #1bddce6

modules/Cockpit/assets/components/field-select.tag
piotr-cz: Here I'm converting values from `field-select` back to integer: https://github.com/agentejo/cockpit/pull/1246/files#diff-9f7295b64abd3cc5e2d21b9b01b75af2R302-R305

It doesn't feel completely correct, I admit

piotr-cz: If you'd like to keep strict value comparison in field-select, another option is to convert direction to string specially for the field-select and than cast back, as here: https://github.com/piotr-cz/cockpit/commit/b5a481003619d8653ce84c43313dea04792c6e49

Maybe there is better way but I'm not familiar with Riot.js

aheinze:
Thanks!

PR #1247, opened 2020-02-13
Fix URL generation when site_url and root paths are not the same

21stcaveman:
In my setup, I use the defines.php file to define COCKPIT_SITE_DIR , COCKPIT_STORAGE_FOLDER and COCKPIT_PUBLIC_STORAGE_FOLDER separately for each site, based on EnvVars set in apache config.
When doing that, URLs generated for AssetManager generated javascript, and any Image in Finder were messed up, so none of them could be previewed or opened in a separate tab.
This commit aims to fix that, with having the original code and setup in mind.

I have tested this with the original setup (following normal cockpit install instructions, with no changes and no defines.php), and with my new setup.

### DIFF
--- a/lib/Lime/App.php
+++ b/lib/Lime/App.php
@@ -420,7 +420,7 @@ class App implements \ArrayAccess {
             $url = \implode('/', \array_map('rawurlencode', explode('/', $url)));
 
             if ($full) {
-                $url = \rtrim($this->registry['site_url'], '/').$url;
+                $url = str_replace($this->path('site:'), '/', \rtrim($this->registry['site_url'], '/').$url);
             }
         }
 
--- a/lib/LimeExtra/App.php
+++ b/lib/LimeExtra/App.php
@@ -192,7 +192,7 @@ class App extends \Lime\App {
                 $this->helper('fs')->write($path, implode("\n", $contents));
             }
 
-            $url = $this->pathToUrl($path);
+            $url = $this->pathToUrl($path, ($this->path('site:') != $this->path('#root:')));
             $list[] = '<script src="'.($url.($version ? "?ver={$version}":'')).'" type="text/javascript"></script>';
         }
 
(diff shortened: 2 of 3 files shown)


aheinze:
di you also test a default cockpit setup in a subfolder of the documentroot?

21stcaveman:
> di you also test a default cockpit setup in a subfolder of the documentroot?

Yes, that's what I reffered to by the original setup.

aheinze:
Sorry, I can't merge the PR because you're using hardcoded path namespaces in the Lime\App class

eg `$this->path('site:')` 

I've updated the `pathToUrl` method, can you please test the current next branch?

21stcaveman:
Unfotunately, that change does not address the problem. Let me explain a bit more. Consider :
COCKPIT_DIR = /var/www/admin/
COCKPIT_SITE_DIR = /var/www/test/
COCKPIT_STORAGE_FOLDER = /var/www/test/storage/
base_url = https://admin.test.com
site_url = https://test.com

When generating a full URL, links should look like 'https://test.com/storage/... , where is, with the original code, as well as the new proposed code you referred to, the resulting URL would be 'https://test.com/var/www/test/storage/...'
My change to the 'pathToUrl' function, aimed to remove the extra COCKPIT_SITE_DIR from the generated URL, only if full URL was requested.

Also, please consider the changes to Media.php as well as LimeExtra/App.php. If full URL is not requested in those locations, the url returned would be 'https://admin.test.com/var/www/test/storage/...', as a result of  'base_url' and 'site_url' being different.

Can you tell me which key in app->registry holds COCKPIT_SITE_DIR and COCKPIT_DIR? I can fix the code based on that.

aheinze:
This is a setup cockpit was never designed for. Only if cockpit is eg in a subfolder of the main project - `https://test.com/cockpit`

need to think about it. But thanks for coming up with this.

21stcaveman:
Understood, I will close this PR.

Absolutely. This setup, allows an admin to use cockpit as the admin panel for several websites at the same time (on the same server), and makes it easier to maintain.

21stcaveman:
Also, in case this is of interest at any point, I believe replacing `$this->path('site:')` with `$this->paths('site')[0]` would address the concern you had about hardcoded path namespaces in LimeApp. Please see my fork for appropriate fixes to make this work with the "centralized" setup as described above.

PR #1249, opened 2020-02-18
Localized collection-link with better display

fredrikkarlsson-pervanovo:
* Added language selector to collection-link and caching display on a key composed of link.id and lang.
* Collection-link display option can now handle an array of fieldNames and returns a comma separated display string


aheinze:
Thanks for the contribution! Will review it soon!

fredrikkarlsson-pervanovo:
Great! If there is any documentation I can update with the display field, point me in the right direction :)

PR #1251, opened 2020-02-20
Read-Only UI Feature

markuslueckingmuc:
Added Readonly-UI feature to the following Fields:

- Boolean
- Date
- Location
- Text
- Time
- CollectionLink

It's only an UI-Feature


aheinze:
why don't you use the access configuration in the field settings? 🤔

aheinze:
closing this...all field options do accept a `disabled` setting. but thanks for the contribution!

Issue #1082, opened 2019-04-14
Kubernetes


jbonnett92:
Please can you add Kubernetes support?

sogeniusio:
+1

jnsvd:
+1 (Willing to contribute on this)

aheinze:
Sorry, but I'm totally unexperienced with Kubernetes. What is missing?

jnsvd:
At the moment there is only Docker support, which is not sufficient if you want to integrate the CMS in a managed containerised environment, i.e. Kubernetes. Integrating in CI/CD pipelines would also require a bit of additional work.

For example:
* You need to call `/install` endpoint before you can login. If Kubernetes schedules a pod running the container, it should be ready for usage once it's running. All initialisation should be done without requiring to call an endpoint.
* Configuration is now done through PHP files and not through yaml files. You want to define a `ConfigMap` containing all allowed configurations, and read those in the code. You also want to allow (some) changes at run-time.
* A default username and password should be configurable through Kubernetes `Secret`
* ...

Most of these "missing" things can be implemented by anyone that wants to set it up in Kubernetes, but it would be more convenient to have it wrapped in a package. Ideally a Helm chart for Cockpit would exist here: https://github.com/helm/charts/tree/master/stable

jdbohrman:
I'd love to help on this as well!

jdbohrman:
www.codefresh.com is the ticker for CI/CD.

jdbohrman:
> At the moment there is only Docker support, which is not sufficient if you want to integrate the CMS in a managed containerised environment, i.e. Kubernetes.

Yeah this is definitely true. 

> Integrating in CI/CD pipelines would also require a bit of additional work. This isn't completely true, you should check out Codefresh, it would really save you some time.

jbonnett92:
This does not require CI/CD but rather a template YAML file made for integrating Kubernetes

mpern:
It's "just" a matter of providing a sane default `config.php` that reads everything from environment variables, and of course, removing the requirement to hit `/install` (there should be a way to either provide the admin password for the bootstrap, or you auto-generate a safe one during startup)

With this in place, writing a helm chart is possible.

abenelazar:
I would also like to help with this! I'm lacking the PHP expertise to make it happen, but I tried on my own for a bit.

mpern:
Maybe it would also make sense to add capabilities to load parts of the config from files to avoid leaking secrets to the environment variables

[...]

Issue #619, opened 2017-12-11
Specify admin user


Kamshak:
Hey,

I really like the way i can just do docker run -d --name cockpit -p 8080:80 agentejo/cockpit as you indicate on the website. Only problem: I have no idea what user to put in. Kind of ruins the nice onboarding flow.

Why not add a -e ADMIN_USER=admin -e ADMIN_PASSWORD=password or something similar for getting started quickly. Bettter yet, a demo page :)

Kamshak:
OK i figured it out :) 
Please add a quick line on the homepage to navigate to /install after launching the container

jordanboston:
I think it shows that right here on the setup in the docs: https://getcockpit.com/documentation/getting-started/installation
Under **Create your first user.**

Kamshak:
@aheinze would it be possible to put this into https://getcockpit.com/download below the docker cmdline? Or link to the docs? Would remove some friction from the initial UX.

b01:
So yeah, my eyes jumped past the line "Go to /url/to/cockpit/install in your favorite Browser and follow the instructions." and when straight to the header "Installation using Docker" so I missed that. Would be nice if the first run just takes you to the install, or update the docker doc section to say go to that URL first.

Kamshak:
![img](http://i.imgur.com/sutiDlW.png)

This page is amazing, it's really appealing to just put that command in the command line. Unfortunately after running it you have no idea what to do :smile: It would be soooo great to have that line to navigate to the install page :)

I love this project, just trying to help here :)

rm-rf-etc:
@Kamshak is exactly right

kiwimonk:
This is still a problem

asarra:
Can second that. Still a problem.

beasteers:
+1 I agree that user/pwd via environment variables would be really convenient

petersg83:
I launched Cockpit via Docker, it's been 10 minutes I'm looking what are the default username & password to log in. Still haven't found it.
Would be nice to write what are the default username & password at least on the docker hub page

PR #1257, opened 2020-02-24
Admin user autoinstall - based on environment variables

beasteers:
#### The issue
Previously, on first run, you'd have to go to `/install` before the instance starts working. 
This is in reference to: https://github.com/agentejo/cockpit/issues/1082 and https://github.com/agentejo/cockpit/issues/619

#### A solution
Now, you can specify admin user options via environment variables and it will create the admin user automatically on first run. Any omitted variables will use the original default options.

#### An example:
```yaml
version: '3.3'

services:
  cockpit-test:
    image: agentejo/cockpit:autoinstall
    container_name: cockpit
    ports:
      - "8080:80"
    volumes:
      - ../github/cockpit:/var/www/html/
      - ./storage:/var/www/html/storage
    environment:
      - COCKPIT_ADMIN_USER=asdf
      - COCKPIT_ADMIN_NAME=Asdf Asdf
      - COCKPIT_ADMIN_PASSWORD=asdf
      - COCKPIT_ADMIN_EMAIL=asdf@asdf.asfd
      - COCKPIT_ADMIN_I18N=en
      - COCKPIT_SALT=asddfasdfasdf
```
-----
#### More details
It will only autoinstall if `COCKPIT_ADMIN_PASSWORD` is defined. If you comment that out, you'll need to go to `/install` like before, but the other parameters will still take effect (your admin username would be `asdf` with a password `admin` like before and a message saying to change it).

After autoinstalling, it unsets the `COCKPIT_ADMIN_PASSWORD` environment variable (1. so that it's not floating around needlessly, and 2. so that the install will be entirely skipped subsequently).

#### Further discussion
There was a discussion about using a `config.php` file in the Kubernetes thread. I'm not opposed to that as an alternative option, but personally, environment variables are more convenient because I always run it in docker/compose and the security can be handled externally, and just passed in as an environment variable.

So as an approach for the config file, the default could look like:
```php
$COCKPIT_ADMIN_USER = getenv('COCKPIT_ADMIN_USER')
$COCKPIT_ADMIN_PASSWORD = getenv('COCKPIT_ADMIN_PASSWORD')
...
```
and you just overwrite the file with your own hardcoded configuration if you so wish.

### DIFF
--- a/modules/Collections/Controller/Admin.php
+++ b/modules/Collections/Controller/Admin.php
@@ -517,25 +517,13 @@ class Admin extends \Cockpit\AuthController {
 
     protected function _filter($filter, $collection, $lang = null) {
 
-        if ($this->app->storage->type == 'mongolite') {
-            return $this->_filterMongo($filter, $collection, $lang, true);
-        }
-
-        if ($this->app->storage->type == 'mongodb') {
-            return $this->_filterMongo($filter, $collection, $lang);
-        }
-
-        return null;
-
-    }
-
-    protected function _filterMongo($filter, $collection, $lang, $isMongoLite = false) {
+        $isMongoLite  = ($this->app->storage->type == 'mongolite');
 
         $allowedtypes = ['text','longtext','boolean','select','html','wysiwyg','markdown','code'];
         $criterias    = [];
         $_filter      = null;
 
-        $this->app->trigger("collections.admin._filterMongo.before", [&$filter, &$allowedtypes, &$criterias]);
+        $this->app->trigger('collections.admin._filter.before', [$collection, &$filter, &$allowedtypes, &$criterias]);
 
         foreach ($collection['fields'] as $field) {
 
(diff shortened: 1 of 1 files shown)


beasteers:
@aheinze Hey just thought I'd bump this since it's a relatively small fix that is pretty helpful and doesn't break existing functionality.

beasteers:
So I just tested this again and the install was throwing errors, presumably due to some change in the core - maybe db initialization or something, I'm not sure, but I'm guessing that something is happening in `apache2_foreground` that needs to happen before any database stuff can be done.

##### Before
to run the install I had just added an install script to the Dockerfile CMD: `php -f install/init.php` and that worked fine back in February. But now that no longer works. 

##### Now
in `index.php` we:
 - check if `COCKPIT_ADMIN_PASSWORD` is set, 
 - then we check if there are any users in the database, 
 - if not then we do the regular install checks that you do in `install/index.php`, (is writable, etc)
 - and add the admin user based on the configured environment variables.

I'd like to avoid having the install checks in `index.php` and only run on startup, but I'm not sure how to do that.

----
But so people know, this works now and I've pushed up a docker image for this to use until someone looks at this PR.
https://hub.docker.com/r/beasteers/cockpit

beasteers:
I was dumb when I opened this and used the default branch so I'll need to re-open this with a new branch, but at the rate that it's been going at I don't forsee it being too urgent lol

PR #1266, opened 2020-02-27
Escape App.Utils.renderer.default()

raffaelj:
Escape `<` and `>` in `App.Utils.renderer.default()` to prevent wrong opened/closed html elements inside raw field.

It seems, that other characters doesn't need to be escaped, because riot.js does that jpb internally.

It would escape wysiwyg contents, so I added a simple wysiwyg renderer, that truncates words.


aheinze:
This is the wrong approach, you should do the escaping in `App.Utils.renderer.wysiwyg`  
`App.Utils.renderer.default` should stay as it is

raffaelj:
The wysiwyg field didn't need escaping, but I left the renderer with the truncated word string with stripped tags.

I also reverted the change to the default renderer and added a text and a textarea renderer instead. Now they both are escaped.

Before escaping the text field:
cp-renderer-specialchars-before

After escaping:
cp-renderer-specialchars-after

aheinze:
Thanks Raffael!

PR #1274, opened 2020-03-15
Add Cockpit + PHP7.2 + Nginx docker image

gabskoro:
I created a docker image that: 
* Used PHP 7.2 + required modules (sqlite3, gd, mbstring, curl)
* Uses Nginx
* Downloads the application zip file  from this repository

If you think this is nice to have, feel free to merge it or let me know if we should maybe move the text somewhere else rather then on the Readme page

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,11 @@ make also sure that <code>$_SERVER['DOCUMENT_ROOT']</code> exists and is set cor
 3. Go to __/cockpit/install__ via Browser
 4. You're ready to use Cockpit :-)
 
+or
+
+1. Run the command `docker run -p 8080:80 gabskoro/cockpit-nginx`
+2. Open `http://localhost:8080/install`
+
 
 ### Build (Only if you modify JS components)
 


mpartipilo:
This is a strange pull request, just changing the readme to promote your own image.

Why not contribute the Dockerfile to this repo instead?

The current one uses Apache, another one replacing Apache for Nginx would be rather nice.

gabskoro:
C'mon guys, really... I didn't even one minute think about it as self-promotion. That's why I also wrote that we can move it somewhere else. Also if you guys have another idea I'm fine with it.

How would I contribute to the Dockerfile, maybe by adding a new one calling `Dockerfile.nginx`?

mpartipilo:
No worries, it was just strange for me 😄 

I would favor Dockerfile-nginx over Dockerfile.nginx, but that's a personal choice.

Let's see what other peers come up with.

aheinze:
@gabskoro thanks for the contribution! is there a reason why you're using php 7.2 and not eg 7.4?

gabskoro:
@aheinze because there is no 7.3/7.4 version available through the debian repo, probably there is a way to add it, but I didn't spend any time looking for it. I also now see that they officially support it till 30 November 2020, so probably would make sense to investigate it

PR #1288, opened 2020-04-03
handles listing and upload based on field configuration

pauloamgomes:
This should not break existing functionality.

Having an asset field configured like:

```
{
  "pattern": "*.png|*.jpg|*.pdf"
}
```

When opening the assets dialog it will only list files matching the pattern.
When trying to upload an asset it will fail with an alert (UIKit uploadSelect) if file extension doesn't match the pattern.


aheinze:
Thanks for the contribution! I'll try to check the PR soon!

PR #1295, opened 2020-04-23
Merging parentComponent options growing exponentially

fredrikkarlsson-pervanovo:
I don't really know why, but merging parentComponent options grows exponentially making the entry page very slow to load and possibly out of memory when you have many layout components.
By debugging and testing copying data in different ways, I found out that cloning the data before extending helped.

This fixes slow entry pages with many layout components. Not sure how many has had problems with it.

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -2783,7 +2783,7 @@ riot.tag2('field-layout', '<div class="uk-sortable layout-components {!items.len
         }
 
         if (opts.parentComponent && opts.parentComponent.options) {
-            opts = App.$.extend(true, opts.parentComponent.options, opts);
+            opts = App.$.extend(true, {}, opts.parentComponent.options, opts);
         }
 
         this.on('mount', function() {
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks for the contribution! Need to take a deeper look into the issue first.

modules/Cockpit/assets/components.js
```diff
@@ -2783,7 +2783,8 @@ riot.tag2('field-layout', '<div class="uk-sortable layout-components {!items.len
         }
 
         if (opts.parentComponent && opts.parentComponent.options) {
-            opts = App.$.extend(true, opts.parentComponent.options, opts);
+            var clonedParentComponentOptions = App.$.extend(true, {}, opts.parentComponent.options);
+            opts = App.$.extend(true, clonedParentComponentOptions, opts);
```
aheinze, line 2787:
```
opts = App.$.extend(true, {}, opts.parentComponent.options, opts);
```

so there is no reason for clonedParentComponentOptions

fredrikkarlsson-pervanovo: Maybe, I will try it on a entry I have problems with

fredrikkarlsson-pervanovo:
@aheinze Implemented your suggestion, seems to work out!

aheinze:
Thanks!

Issue #937, opened 2018-11-22
[Request] Add file type filter to Asset field


andrevvm:
It would be nice to be able to limit Asset fields by file type, either by MIME, extension, or categorized type (i.e. [Kirby CMS](https://getkirby.com/docs/cheatsheet/file/type)).

pauloamgomes:
Think that makes all sense, currently Cockpit permits to limit the upload (global) based on a whitelist rule in the configuration. However, that is not practical for some cases, e.g.:

- A document field would be probably limited to PDF files
- An image field would be limited to only common web formats

Such configuration should reside in the field definition, e.g.:

```
{
  "pattern": "*.png|*.jpg|*.pdf"
}
```

PR #1300, opened 2020-05-15
Support restricting mimetype in Asset fieldtype with "typefilter" option

paulthewalton:
Related to issue #937.
Should be compatible with pull request #1288, but also may overlap some functionality

- passes "typefilter" option through to cp-assets component
- validates it as one of the pre-existing typefilters: "image", "video", "audio", "document", "archive", or "code"
- defaults to empty string (no filter)

Copied from my comment below:

> Not sure what the browser compatibility targets are for this project, so I've tried to keep it as wide as possible.
>
> 
> #### Usage:
> In Asset fieldtype options:
> 
> {
>   "typefilter": "image"
> }
> Setting this option will function the same as picking the typefilter from the dropdown in the UI, but the user will be unable to change the filter. Anything other than one of the predefined typefilter names (case-insensitive) will be discarded, and filtering will work as usual.


aheinze:
Thanks for the contribution and initializing to provide a solution for it. Fixed in https://github.com/agentejo/cockpit/commit/e281ffefb66a2e0cf5b99fa3a5e5eb55b94a9f32

PR #1312, opened 2020-06-18
allow relative urls for content preview

### LINKED ISSUE #1311: content preview input field doesn't allow relative urls
I want to use a relative url `/livePreview` as content preview url and the collection settings page complains about it because it has an `input type="url"`. If I change it to `type="text"` in my browser or if I change the `pages.collection.php` manually, my content preview works as expected.

But now I'm not able to save changes via admin ui because the preview input field complains in a not selected/not visible tab without a visible error message. If I change it back to `type="text"`, I can work with it.

The reason for using a relative url is to keep my local dev setup in sync with the production server without having to change `http://localhost/livePreview` to `https://example.com/livePreview` all the time.

raffaelj:
fixes #1311


aheinze:
Hi! Thanks for the contribution and the hint for improvement! 

I've come up in the latest commit with another approach:

Use `site://` or `base://` to get relative urls for your cockpit setup.

Greets
Artur

PR #1316, opened 2020-07-09
Configurable templates

raffaelj:
Singletons now have the same template functionality like collections, so existing singletons are proposed when hovering over "Add field" while creating a new singleton.

Collections and singletons can have custom templates in `#config:collections/templates/*.php` and `#config:singletons/templates/*.php`.

For collections: If the custom config templates folder exists, the folder `collections:fields-templates` will be ignored.

### DIFF
--- a/modules/Collections/Controller/Admin.php
+++ b/modules/Collections/Controller/Admin.php
@@ -105,7 +105,12 @@ class Admin extends \Cockpit\AuthController {
         // get field templates
         $templates = [];
 
-        foreach ($this->app->helper('fs')->ls('*.php', 'collections:fields-templates') as $file) {
+        $templatesPath = 'collections:fields-templates';
+        if ($customTemplatesPath = $this->path('#config:collections/templates')) {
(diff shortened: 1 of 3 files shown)


modules/Collections/Controller/Admin.php
```diff
@@ -105,7 +105,12 @@ public function collection($name = null) {
         // get field templates
         $templates = [];
 
-        foreach ($this->app->helper('fs')->ls('*.php', 'collections:fields-templates') as $file) {
+        $templatesPath = 'collections:fields-templates';
+        if ($customTemplatesPath = $this->path('#config:collections/templates')) {
```
aheinze, line 109: why not allow both? collect the paths in an array and then do a loop over the folders

raffaelj: This was intentional to avoid naming conflicts and to have the ability to override the default template(s). If I choose to use custom templates I can still copy the core collection template to my template folder.

PR #1317, opened 2020-07-11
create bootstrap.php when creating folder via finder in addons dir

### LINKED ISSUE #1215: Cockpit breaks on empty addons folder
I tried uploading an addon using Finder, but the moment I created a folder within the addons folder, the entire CMS stopped working, with the server responding with a 500 http status code.

Apparently, Cockpit will try to load any folder found within addons as an Addon, without checking the validity of the folder's contents. Cockpit should be more defensive when checking for addons, as this prevents the user from being able to upload addons using finder

### LINKED ISSUE #1314: Fatal error: require(): Failed opening required '/var/www/html/addons/icons/bootstrap.php'
Hi, I installed Cockpit using [Docker](https://getcockpit.com/documentation/reference/docker).

Everything was going smooth until I tried creating a folder in the Media Library. When I tried to save I got this error 

```
Warning: require(/var/www/html/addons/icons/bootstrap.php): failed to open stream: No such file or directory in /var/www/html/lib/Lime/App.php on line 1150

Fatal error: require(): Failed opening required '/var/www/html/addons/icons/bootstrap.php' (include_path='.:/usr/local/lib/php') in /var/www/html/lib/Lime/App.php on line 1150

```

Re-started the container to see if it would clear but no luck.

raffaelj:
This issue came up so often and it is still a way to break the whole cockpit instance with a simple folder creation via ui.

Now the media api controller takes care of this and creates an empty `bootstrap.php` file after creating an empty folder inside the addons or the modules folder.

fixes #1215, fixes #1314

### DIFF
--- a/modules/Cockpit/Controller/Media.php
+++ b/modules/Cockpit/Controller/Media.php
@@ -193,7 +193,14 @@ class Media extends \Cockpit\AuthController {
         $ret  = false;
 
         if ($name && $path) {
-            $ret = mkdir($this->root.'/'.trim($path, '/').'/'.$name);
+            $parentFolder = $this->root.'/'.trim($path, '/').'/';
+            $ret = mkdir($parentFolder.$name);
+
+            // prevent fatal error when creating folders without bootstrap.php in addons/modules dirs
+            // see this discussion, why it is necessary; https://github.com/agentejo/cockpit/pull/1019
+            if ($ret && in_array($parentFolder,[$this->path('#addons:'), $this->path('#modules:')])) {
+                @file_put_contents($parentFolder.$name.'/bootstrap.php', "<?php\r\n");
+            }
         }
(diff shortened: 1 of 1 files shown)


aheinze:
I've decided to use a `is_file` check which seems to be more performant than `file_exists`. still need to investigate the impact in the future and eg only do the check in debug mode

raffaelj:
If you decide to do the check only in debug mode, than my fix would still be very useful. If I encounter that bug while developing, I'm able to check the issues or the forum, but if a client or a less advanced developer/designer breaks the installation while not in debug mode, the whole story starts over again.

aheinze:
I'll keep the check for now ... I think in the future we can overcome performance concerns  by using the preloading directive in production introduced in php 7.4

Issue #1323, opened 2020-08-26
collection names aren't quoted in MongoLite


raffaelj:
I renamed a collection from "projects" to "references" via cli and suddenly the entries couldn't be fetched anymore.

Error from `/collections/find` in backend:

```json
{"error":"Call to a member function fetchAll() on bool","file":"\/var\/www\/html\/cockpit\/lib\/MongoLite\/Cursor.php","line":198}
```

Error from CpMultiplane frontend

```text
<b>Fatal error</b>:  Uncaught Error: Call to a member function fetchAll() on bool in /var/www/html/cockpit/lib/MongoLite/Cursor.php:198
Stack trace:
#0 /var/www/html/cockpit/lib/MongoLite/Cursor.php(159): MongoLite\Cursor-&gt;getData()
#1 /var/www/html/cockpit/lib/MongoHybrid/MongoLite.php(100): MongoLite\Cursor-&gt;toArray()
#2 [internal function]: MongoHybrid\MongoLite-&gt;find('collections/ref...', Array)
#3 /var/www/html/cockpit/lib/MongoHybrid/Client.php(425): call_user_func_array(Array, Array)
#4 /var/www/html/cockpit/modules/Collections/bootstrap.php(248): MongoHybrid\Client-&gt;__call('find', Array)
#5 /var/www/html/cockpit/lib/Lime/App.php(1260): Lime\Module-&gt;{closure}('references', Array)
#6 /var/www/html/modules/Multiplane/bootstrap.php(637): Lime\Module-&gt;__call('find', Array)
#7 /var/www/html/cockpit/lib/Lime/App.php(1260): Lime\Module-&gt;{closure}(Array, 'references', Array)
#8 /var/www/html/modules/Multiplane/Controller/Base.php(34): Lime\Module-&gt;__call('getPosts', Array)
#9 /var/www/html/cockpit/lib/Lime/App.php(967): Multiplane in <b>/var/www/html/cockpit/lib/MongoLite/Cursor.php</b> on line <b>198</b><br />
```

I digged a bit and the solution was simple - and it's actually impressive, that it didn't happen earlier. I guess, that this issue wasn't possible before [dropping unique ids for collection _id](https://github.com/agentejo/cockpit/commit/c374927cf9417cdd8414c35e564f298c2bf6b8a1#diff-aabc3c53f30f46e10259403de9fff101).

Now it's possible to use reserved words by accident (e. g. "references"): https://www.sqlite.org/lang_keywords.html

[...]

PR #1325, opened 2020-08-30
more quotes for collection names (MongoLite)

raffaelj:
Same issue as #1323, but now it happened when I updated an entry of a collection named "references" via admin ui

### DIFF
--- a/lib/MongoLite/Collection.php
+++ b/lib/MongoLite/Collection.php
@@ -106,7 +106,7 @@ class Collection {
         $fields = \implode(',', $fields);
         $values = \implode(',', $values);
 
-        $sql = "INSERT INTO {$table} ({$fields}) VALUES ({$values})";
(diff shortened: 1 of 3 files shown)


lib/MongoLite/Collection.php
```diff
@@ -224,7 +226,7 @@ public function renameCollection($newname) {
 
         if (!in_array($newname, $this->database->getCollectionNames())) {
 
-            $this->database->connection->exec('ALTER TABLE '.$this->name.' RENAME TO '.$newname);
+            $this->database->connection->exec('ALTER TABLE '.$this->database->connection->quote($this->name).' RENAME TO '.$newname);
```
aheinze, line 229: shouldn't `$newname` also be "quoted"?

aheinze (changes requested):

aheinze:
Thank you 🙏

PR #1326, opened 2020-08-31
pass missing opts in cp-field tag

raffaelj:
The opts `cls` and `required` got lost somehow in the past. Now text fields can have a custom class again and also the browser validation fires if required fields are empty before the form submits.

See: https://discourse.getcockpit.com/t/trying-to-apply-custom-class-to-text-field/1604

I did some minor cleanup, too.


aheinze:
Thanks!

PR #1349, opened 2020-10-12
Field layout preview component trigger

fredrikkarlsson-pervanovo:
This trigger is needed to be able to write addons that generates layout component previews for custom components

### DIFF
--- a/modules/Cockpit/assets/components/field-layout.tag
+++ b/modules/Cockpit/assets/components/field-layout.tag
@@ -39,9 +39,9 @@
         }
 
         .layout-field-preview canvas {
-            background-size: contain; 
-            background-position: 50% 50%; 
-            background-repeat: no-repeat; 
+            background-size: contain;
+            background-position: 50% 50%;
+            background-repeat: no-repeat;
         }
 
         .layout-field-preview:empty {
@@ -265,7 +265,7 @@
             }
         }
 
-        
+
         this.on('mount', function() {
 
             this.showPreview = opts.preview === undefined ? true : opts.preview;
@@ -522,9 +522,15 @@
         }
 
         getPreview(component) {
-            var triggerOutput = [];
-            App.trigger('field.layout.component.preview', {component: component, def: this.components[component.component], output: triggerOutput});
-            if (triggerOutput.length > 0) return triggerOutput.join('');
+            var params = {
+                component: component,
+                definition: this.components[component.component],
+                output: false
+            };
+
+            App.trigger('field.layout.component.preview', params);
+
+            if (params.output) return params.output;
 
             var def = this.components[component.component];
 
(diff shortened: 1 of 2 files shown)


modules/Cockpit/assets/components/field-layout.tag
```diff
@@ -522,7 +522,9 @@
         }
 
         getPreview(component) {
-            //console.log(component)
+            var triggerOutput = [];
+            App.trigger('field.layout.component.preview', {component: component, def: this.components[component.component], output: triggerOutput});
+            if (triggerOutput.length > 0) return triggerOutput.join('');
```
aheinze, line 527: Why an array?

fredrikkarlsson-pervanovo: For the handler to be able to send back data it has to be some kind of referenceable data type, ie. object or array. A string would not be able to be changed from inside the handler.

Here is some code of the handler in our addon that uses this trigger:
```
 App.on("field.layout.component.preview", function (trigger) {
        var componentName = trigger.params.component.component;
        var value = trigger.params.component.settings;
        var output = trigger.params.output;
        switch (componentName) {
            case "custom_heading":
                output.push(badge(value.tag) + value.text);
                return;
```

fredrikkarlsson-pervanovo: @aheinze Any toughts on this? I just rebased the PR to stay in sync.

fredrikkarlsson-pervanovo:
Here is an addon I just published that needs this PR: https://github.com/Pervanovo/CockpitLayoutComponentPreviews

modules/Cockpit/assets/components/field-layout.tag
aheinze:
```js
var output = '',
    params = {
        component: component, 
        def: this.components[component.component],
        render: function(contents) {
            output = contents || '';
        }
    };

App.trigger('field.layout.component.preview', params);

if (output) return output;
```

your handler would look then like:

```js
 App.on("field.layout.component.preview", function (trigger) {
        var componentName = trigger.params.component.component;
        var value = trigger.params.component.settings;
        var render = trigger.params.render;
        switch (componentName) {
            case "custom_heading":
                render(badge(value.tag) + value.text);
                return;
```

fredrikkarlsson-pervanovo: Great, I will rewrite it!

fredrikkarlsson-pervanovo:
@aheinze Changed according to your feedback!

aheinze:
> @aheinze Changed according to your feedback!
I think you still need to compile it (re-create components.js)

fredrikkarlsson-pervanovo:
> > @aheinze Changed according to your feedback!
> 
> I think you still need to compile it (re-create components.js)

Silly me, I forgot... Done!

aheinze:
Thanks!

PR #1350, opened 2020-10-16
Trigger needed in collection entries for addon development

fredrikkarlsson-pervanovo:
To be able to add functionality with an addon to Collection entries view, a trigger like this is needed.

### DIFF
--- a/modules/Collections/views/partials/entries.php
+++ b/modules/Collections/views/partials/entries.php
@@ -682,4 +682,5 @@
 
     </script>
 
+    @trigger('collections.entries', [$collection])
 </div>


aheinze:
whhy don't you trigger it her: https://github.com/agentejo/cockpit/blob/next/modules/Collections/views/entries.php ?

fredrikkarlsson-pervanovo:
> why don't you trigger it here: https://github.com/agentejo/cockpit/blob/next/modules/Collections/views/entries.php ?

I don't remember anymore, I will try to move it and see if anything breaks

fredrikkarlsson-pervanovo:
I tries to move the trigger, but it broke my event handler, that looks like this:
```
$this->on('collections.entries', function ($collection) use ($app) {
  $filters = $app->module("collectionentriesfilters")->loadFilters();
  $this->renderView("collectionentriesfilters:views/partials/filters.php", ['collection' => $collection, 'filters' => $filters[$collection['name']]]);
});
````
Basically I render more content in the collections.entries handler and then move it with javascript. But `this.mount` does not seem to be called if the trigger is not inside something which is passed thru @render.
How do I go about this?

fredrikkarlsson-pervanovo:
Tried to move it, but I need to have trigger in the partial since I have built a riot-view that modifies the filter field and it needs to be redrawn when the filtering process is run.

Would you accept the PR if I change the trigger name?

aheinze:
Thanks!

PR #1353, opened 2020-10-20
Bugfix field-file preview

fredrikkarlsson-pervanovo:
The field preview on collection entries page for field-file is broken. This PR fixes it.


aheinze:
Thanks!

PR #1354, opened 2020-10-20
Added triggers to collection admin find filters

fredrikkarlsson-pervanovo:
To be able to handle default search functionality in custom fieldtypes we needed triggers where cockpit creates the default search filters.

Also, the mongo and mongolite internal filter functions were similar down to adding the case sensitive options, so I merged them and added a boolean flag to select if it should be for mongolite.

### DIFF
--- a/modules/Collections/Controller/Admin.php
+++ b/modules/Collections/Controller/Admin.php
@@ -535,6 +535,8 @@ class Admin extends \Cockpit\AuthController {
         $criterias    = [];
         $_filter      = null;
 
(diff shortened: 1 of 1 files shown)


aheinze:
Thanks! I'll do some cleanup on this in my next commit, so you just are aware of it!

fredrikkarlsson-pervanovo:
Alright, changing the names of the triggers?

aheinze:
https://github.com/agentejo/cockpit/commit/74c0eb527bda655b6f7e2f87201a3824b3f56c00

fredrikkarlsson-pervanovo:
👍

PR #1355, opened 2020-10-20
Added open to field-file dropdown menu

fredrikkarlsson-pervanovo:
Our editors feel limited that we cannot easily open files linked in file fields. 
This PR adds a "Open in new tab" option to the field-file dropdown menu.


aheinze:
Thanks!

PR #1356, opened 2020-10-21
Localize collectionlink renderer

fredrikkarlsson-pervanovo:
The collectionlink renderer calls find on the linked collection and entry but not with the correct lang, which is not even provided. 
This PR adds lang to App.Utils.renderValue and App.Utils.renderer.collectionlink.


aheinze:
Thanks!

PR #1357, opened 2020-10-21
Persist asset selector folder and filter

fredrikkarlsson-pervanovo:
The finder persists it state and it would be usefull if also the asset selector did the same.


aheinze:
Thanks!

PR #1358, opened 2020-10-22
Bugfix collectionlink renderer

fredrikkarlsson-pervanovo:
Somehow I missed to check that the field preview got updated when switching language. Since it cached the first one and then would use that one for all languages. This PR adds the lang to the cache key.


aheinze:
Thanks!

PR #1359, opened 2020-10-22
Better Add component label

fredrikkarlsson-pervanovo:
When adding a component, it is nice to see what parent it adheres to. Look at the following example:
image
Then it is nice to know what parent will receive the child.
This PR adds the parent component label to the add component (child) button bellow the list of components.


fredrikkarlsson-pervanovo:
Rebased this to be easily merged

PR #1362, opened 2020-10-28
Bugfix updated admin filters

fredrikkarlsson-pervanovo:
The mongolite+mongo merge has a bug. It currently adds the mongo option to the wrong field for collectionlink and location.
This PR adds the mongo options to the correct field. I am sorry that I created this bug :(


aheinze:
Thanks!

PR #1373, opened 2020-11-16
webp support

patrickuhlmann:
add the webp image format to all filter. This allows the assets to be
recognized as images and to have thumbnails been generated. It also
allows webp images to be used in the assets manager.


aheinze:
Thanks!

PR #1374, opened 2020-11-16
better filenames

patrickuhlmann:
before the filename started with a md5 hash and the original name came
afterwards. Now the original name comes first and is then followed by a
uid. I also made the uid explicitly visible with a prefix.

In the assets manager files are displayed based on their filename. With
this change this is improved too as now the sorting is based on the
original filename and not on the md5 hash. This has lead to some strange
ordering before.


aheinze:
Thanks!

PR #1376, opened 2020-11-19
added i18n translation for UIkit password field and a missing placeholder translation

ciruz:
A small fix, the link label translation for the UIkit password field got lost after clicking it.

before:
before

after:
after


aheinze:
Thanks!

PR #1377, opened 2020-11-24
Adds cyrillic slug transliteration

rensite:
Cyrillic slugs are not supported yet. This is a quick fix that transliterates Cyrillic characters into latin.


aheinze:
Спасибо!

Issue #1331, opened 2020-09-13
[Feature] new attribute "protected" on fields


L3444:
In our project we would like to make some fields in our collections "protected", meaning they must be filled in when initially creating the entity but can't be changed afterwards in the UI (or maybe can be changed, but only after accepting a confirm-dialog). This "protected" attribute should probably only be checkable when "required" is already checked. The purpose of this attribute would be to prevent the editors from accidentally changing a vital field of the entity that e.g. is meant to identify this entity.

Issue #677, opened 2018-02-21
[Help] How to hide|read only a field for non admins when editing/creating an entry in a collection


fabianmu:
I do have a `slug` field, which I set automatically via `$app->on("collections.save.before.{collectionName}"` and I would like that field to be either not visible or read-only when a non-admin user (I have created other groups via `groups:..` in `config.yaml` is creating or editing an entry.  

I saw a comment earlier about custom PHP code in a collections *Permissions*, #676 which may be useful but I need some guidance on if and how it's possible to adjust an entry to accomplish that desired outcome.
Thanks!

pauloamgomes:
I believe that requires a read-only or hidden attribute on the field definition so we can define the field like:

```
{
  "hidden": true
}
``` 

The hidden field can be added easily to the input:

```
if (opts.hidden) {
    App.$(this.refs.input).addClass('uk-hidden');
}
```

but believe that should be performed in a more generic way and maybe including other field types than text.

The readonly option seems a bit more simple as it can be handled as new attribute:

```
{
  "readonly": true
}
``` 

and the text template can be updated to:

```
(['disabled', 'maxlength', 'minlength', 'step', 'placeholder', 'pattern', 'size']).forEach( function(key) {
    if (opts[key]) $this.refs.input.setAttribute(key, opts[key]);
});
```

think is a small change that provides the flexibility you described above

adambuczek:
This is probably a common case.  
I was under the impression that Field's Access tab influences not only Backend but also API response but no matter which API key I use the data returned doesn't have `slug` field.  
Is this by design? Maybe this would be the way to go?

BTW I would also like to know more about what can I do in collection settings so I could implement this or e.g. collection level id filter myself.

michabbb:
hi, i have the same problem. how can i hide or make a field read-only ? i guess a primary-key is something very common and never should be editable by any user. thanks !

michabbb:
i took a while till i found out, that i have to make these changes in   
`modules/Cockpit/assets/components.js`. can someone please explain me, what the file 
`modules/Cockpit/assets/components/field-text.tag` is about ? changes here do not have any effect. 
thanks.

adambuczek:
[You need to rebuild js components after you edit them. ](https://github.com/agentejo/cockpit#build-only-if-you-modify-js-components)

raffaelj:
If you want to hide the slug field (and optional more fields) completely for non-admins, you can use this:

```php
<?php

/*
[...]

PR #1389, opened 2021-01-09
Add support for Read Only Fields Based on User Group

jnterry:
# Summary of Changes
- Add `acl_ro` (access control list read only) parameter to fields within a collection
- Displays accessible read-only fields in entries table and entry editor
- Prevents interaction with read only fields by applying `uk-disabled` css class
- Add server side validation on collection/save_entry endpoint to ensure user cannot edit fields they do not have access to by manipulating the JSON payload sent to server

# Motivation

Currently, the per field acl in a collection can be used to hide fields from certain groups of users. For our use case of cockpit, we would like to mark some basic fields about an entry (such as name, slug, etc) as only editable by admins, but allow volenteers to contribute to content within other fields of an entry.

Marking the "basic" fields as only accessible to admins completely hides them from the UI, meaning volenteers cannot see which entry they are editing.

This PR adds seperate access control lists for read/write (the current ACL) and read only - allowing users to see, but not modify fields marked as read only.

# Implementation

## Split ACL

The per-field access control list is now split into two parts:

access-control-lists

The read write list is stored with the `acl` parameter as before, ensuring backwards compatbility with existing cockpit deployments. The new readonly list is stored in a `acl_ro` parameter

## Entries Table

Entries table now displays columns for any field the user can see based on the union of `acl` and `acl_ro`. 

The "bulk edit" interface on the entires table uses only the `acl` list to avoid allowing bulk edits of fields which the user can only see in a read-only capacity.

(this would be a good use case for the upcoming [HTML "inert" attribute ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert)- but since it is not yet supported by any browser (without enabling experimental support), I've used custom javascript. An alternative would be to add the ["inert" polyfill](https://github.com/WICG/inert) to cockpit and use that attribute for a more robust solution.

## Entry Editor

The `disabled` option is passed into the `cp-field` for readonly fields to render them partially translucent and prevent cursor interaction.

I have also added code to prevent a user gaining focus into a disabled `cp-field` by pressing tab on the keyboard to move between inputs.

## Server Side Validation

The existing ACL is not validated server side, meaning a user can reply a http request to collection/save_entry, modifying the JSON payload to overwrite a field they do not have access to.

This PR additionally adds server side validation to the Collection's Admin controller to use the old value for any read only fields.


jnterry:
While this PR was developed for my own use of cockpit, looking at open issues, it may help as a base towards solving:

- https://github.com/agentejo/cockpit/issues/677
- https://github.com/agentejo/cockpit/issues/1331

Issue #1232, opened 2020-01-29
Empty string returned instead of null


luke-lewandowski:
Hi all,

I'm using agentejo/cockpit:0.9.3

The issue I have ran into is that when my collection has a field which is eg. collectionlink that's optional. If nothing has been assigned to the property. 
The API returns empty string "" instead of null which is not strictly accurate. 

Is this something that's done on purpose of a bug that I should investigate and submit a pull request for?

luke-lewandowski:
I'm thinking somewhere eg. here - https://github.com/agentejo/cockpit/blob/3f5d004e15953a73c3c10e12fc78324bd2495df8/modules/Collections/bootstrap.php#L298 there should be case for an object to save as null

Not sure if I'm on the right track as I'm not too familiar with the codebase.

MarcoDaniels:
I've also been struggling with the empty string instead of null problem, so I open a pull request that might fix it, but as described in the PR it will not fix existing data, only newly created one.

To "fix" existing data I've created a very small addon that checks for empty string and defines the return type for each field type on a data request, this might of course have some performance issues as it goes through all the output data on each request.

https://github.com/MarcoDaniels/cockpit-null-check

PR #1393, opened 2021-01-18
Return null instead of empty string

MarcoDaniels:
When using the api to fetch data, in every field type an empty string is returned instead of null. 

This will only fix newly created content, as existing content will already have the empty string saved.

Solves: #1232


aheinze:
Thanks!

PR #1394, opened 2021-01-18
Set default langauge for "admin" user during app install.

Raruto:
Currently a default language is already set for new accounts created via the admin dashboard:

https://github.com/agentejo/cockpit/blob/f024ee64356d2bb0fbde1a8fbf4ba991fa889c9b/modules/Cockpit/Controller/Accounts.php#L75

_Have a nice Day,
Raruto_


aheinze:
Thanks!

Issue #1298, opened 2020-05-04
.htaccess causing internal server error


Kedmun0:
Hi all,

I hope you can help me. This has been bugging me for days. I have been trying to install cockpit CMS onto my website using the extraction into web root solution but it doesn't seem to be working. I get a server error when I try to install the CMS going though /cockpit/install and then when I disable the .htaccess file, it lets me install it and it gives me the option to login to cockpit using the new admin / admin credentials but it then redirects me to a 404 not found page.

I read that it's because the .htaccess file is disabled but when it is enabled, I get an internal server error

aheinze:
Please investigate http + php log files to get more details. otherwise it is hard to point to the right direction

buzztnt:
I'm having the same issue, when I remove everything but 

`# URL rewrites
<IfModule mod_rewrite.c>

    Options -MultiViews

    RewriteEngine On
    # RewriteBase /

    # Protect some contents
    RewriteRule ^.*/?\.git+ - [F,L]

    RewriteCond %{REQUEST_FILENAME} !-f
    RewriteCond %{REQUEST_FILENAME} !-d
    RewriteRule .* index.php [L]

</IfModule>`

It seems to be working fine. Can anyone shed some light on this?

raffaelj:
solved here: https://discourse.getcockpit.com/t/re-htaccess-file-causes-internal-server-error/1437/3?u=raffaelj

> The reason was, that mod_version wasn’t enabled on a shared GoDaddy host.

> If you have the same problem, either activate mod_version for your apache or if you can’t enable it, you can delete or comment out all the <IfVersion >= 2.4> lines in .htaccess to match your setup.

can be closed now.

wlfnkls:
I got the same issue and commented all ifVersion tags, since mod_version is not existent on my system.
But after installing cockpit successfully, I get a 404 to http://dev-pi.local/cockpit/auth/login?to=/.

Any ideas?

MarcTabaries:
> I got the same issue and commented all ifVersion tags, since mod_version is not existent on my system. But after installing cockpit successfully, I get a 404 to http://dev-pi.local/cockpit/auth/login?to=/.
> 
> Any ideas?

In the root .htaccess file (https://github.com/agentejo/cockpit/blob/next/.htaccess#L38), change line 38 
```
#RewriteBase /
```
by
```
RewriteBase /
```

This should work ;)

Issue #764, opened 2018-05-14
Options -MultiViews : Error 500


nebnes:
When I uncomment Options -MultiViews in .htaccess of cockpit I have no error but I can't save anything. 
And when I comment Options -MultiViews I have an error 500 from the server.

PR #1404, opened 2021-01-24
Improved .htaccess security and readability

Raruto:
**List of changes**

- added comments
- added deny rules for the following files:
   - `composer\.(json|lock)` 
   - `package\.json`
   - `(README|CONTRIBUTING)\.md`
   - `Dockerfile`
   - `LICENSE`
   - "hidden" files and directories (whose names begin with a period)
- removed multiple `<Files>` statement in favor of a single `<Filesmatch>` statement

---

**Additional info**

For those interested in deepening I suggest you start reading one of the following examples:

- [h5bp/.htaccess](https://github.com/h5bp/html5-boilerplate/blob/master/dist/.htaccess) (on which I think this file was initially based)
- [drupal/.htaccess](https://github.com/drupal/drupal/blob/9.2.x/.htaccess)

_Have a nice Day,
Raruto_

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -2,20 +2,20 @@
 # Apache/PHP/Cockpit settings:
 #
 
-# Protect application and system files from being viewed
-<FilesMatch "(\.*|composer\.(json|lock)|package\.json|(README|CONTRIBUTING)\.md|cp|Dockerfile|LICENSE|\.(sqlite|sdb|s3db|db|yaml|yml))$">
-  <IfVersion >= 2.4>
-     Require all denied
-  </IfVersion>
-  <IfVersion < 2.4>
-     Deny from all
-  </IfVersion>
+# Deny access to application and system files from being viewed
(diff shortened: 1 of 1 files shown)


Raruto:
**PS** regarding these two issues:

- https://github.com/agentejo/cockpit/issues/1298#issuecomment-630339782
- https://github.com/agentejo/cockpit/issues/764

---

Some apache (shared) hosts don't allow you to override `Options` directives (thus getting the error 500: `Options not allowed here ...` within server logs).

```py
# sample "httpd.conf" with "AllowOverride" set to "None"

<Directory "/var/www/localhost/htdocs">
    Options Indexes FollowSymLinks MultiViews
    AllowOverride None
    Order allow,deny
    Allow from all
</Directory>
```

As there is no easy way to verify those directives and prevent 500 error (without access to error logs or apache config files), would it make sense to keep them both commented by default?

https://github.com/agentejo/cockpit/blob/722393b31921d5b3ee5992ade63ee9589f5b52a8/.htaccess#L30

https://github.com/agentejo/cockpit/blob/722393b31921d5b3ee5992ade63ee9589f5b52a8/.htaccess#L35

as it happens for the `RewriteBase` directive:

https://github.com/agentejo/cockpit/blob/722393b31921d5b3ee5992ade63ee9589f5b52a8/.htaccess#L38

Issue #1415, opened 2021-02-02
How to translate field in backend


MyTheValentinus:
Hello,

Do you have the possibility to translate field name display for admin user ?

I already use i18n addons, full interface is translate, but not my custom field for collections. I tried to add the field key in my language file but is not worked.

Thanks

MyTheValentinus:
Change in code :

```
<span class="uk-text-bold"><i class="uk-icon-pencil-square uk-margin-small-right"></i> { field.label || App.Utils.ucfirst(field.name) }</span>
#TO ->
<span class="uk-text-bold"><i class="uk-icon-pencil-square uk-margin-small-right"></i> { App.i18n.get(field.label) || field.label } </span>
```

In file `modules/Collections/view/entry.php` line 95

PR #1416, opened 2021-02-02
Add the possibility to translate field label in collections

MyTheValentinus:
Issue #1415 

In this commit only collection field label is edited, but i think it is an good idea to add to other type of items


MyTheValentinus:
It's possible to make translation check exist + upercase ? I have a error in the dev console, if translate not exist, the ucfirst return a error because of null string.
Thanks

PR #1419, opened 2021-02-09
Adds 'mask-type' to allowed SVG attributes

alexksso:
The mask type attribute was missing and it turns out we use it in some elements.


aheinze:
Thanks!

PR #1430, opened 2021-03-19
Allow modified date to be updated with API

MarcoDaniels:
It would be nice to be able to update the `_modified` value with the API. 

This comes handy when migrating data from a different CMS in order to keep the same history.

The param has to be named different from `_modified`, otherwise it will conflict when updating the entry in the backend.


MarcoDaniels:
This doesn't seem like a good idea as the `_modified` date would keep being overwritten.

PR #1431, opened 2021-03-22
Allow filter to be applied at CollectionLinkSelect.

PotterDai:
Allow filter to be applied at CollectionLinkSelect.


aheinze:
Thanks!

PR #1434, opened 2021-04-01
Catch error that is thrown when filter call fails to start

eternalflamez:
This otherwise will block all following calls from starting. This errors often when importing large amounts of data.


aheinze:
Thanks!

PR #1437, opened 2021-04-22
Enhance tags field type

fredrikkarlsson-pervanovo:
We wanted to use the tags field type to share types between different collections and to have autocomplete suggestions from both collections.

To facilitate this, I added the following to field-tags in this PR:
* Loading autocomplete suggestions like select field type (for example to suggest tags used in other entries in same collection)
```
{
  "src": {
    "url": "collection=collection_1",
    "value": "tags"
  }
}
```
* Loading multiple sources (to share tags between collections)
```
{
  "src": [{
    "url": "collection=collection_1",
    "value": "tags"
  }, {
    "url": "collection=collection_2",
    "value": "tags"
  }]
}
```
* Setting minLength option to 0, ie. show autocompletions when focusing the input
* Setting autocomplete delay from option (used to filter options directly when using static list or from sources)
* Don't suggest tag already in use

This PR should be backwards compatible with specifying a URL in the autocomplete option if no sources are used.

### DIFF
--- a/modules/Cockpit/assets/components.js
+++ b/modules/Cockpit/assets/components.js
@@ -4184,37 +4184,99 @@ riot.tag2('field-set', '<div> <div class="uk-alert" if="{fields && !fields.lengt
 
 });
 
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks!

PR #1438, opened 2021-04-22
Print layout-field-preview before children

fredrikkarlsson-pervanovo:
We have added layout component previews for parent elements (ie. having children) and then it makes more sense to have the preview up on top with the heading instead of below the children with the add button.

This PR simply moves the layout component preview area to be printed before the children.


fredrikkarlsson-pervanovo:
Before PR:
image

After PR:
image

fredrikkarlsson-pervanovo:
Rebased this to be easily merged

PR #1447, opened 2021-05-24
Bugfix duplicating localized collection entry

fredrikkarlsson-pervanovo:
When duplicating entries, the currently loaded entry is used (in the selected language) and therefore it writes the localized fields to the default language fields and loses all other localized fields.

This PR mitigates this problem by loading the full entry before saving it and then reloading the entries using current filters.

This bug probably exists in sortable collections as well, but I have not tested that.


aheinze:
Thanks!

PR #1448, opened 2021-06-14
Implemented select renderer

fredrikkarlsson-pervanovo:
Select fields doesn't currently have a special renderer and is therefore just printing it's raw value.
Since select can handle key-value options, it would be nice to present the label for the value if possible.

This PR implements just that, with a tooltip showing the raw value if hoovered:
image


aheinze:
Thanks!

PR #1455, opened 2021-07-12
Allow user to add reply-to option to mailer

### LINKED ISSUE #1453: [Forms] No option for Mailer Reply To
At the moment the only option taken from the Rest API to the Mailer from PHP is the subject, and adding the Reply To would be really nice.

The change is quite small (I was not able to create a PR for it myself, lacking permission for what It looks), the changes would be the following:

In the file `\modules\Forms\Controller\RestApi.php`, submit function, under the scope for adding the __mailsubject option, add the following code:

```php
            if ($this->param('__replyto')) {
                $options['reply_to'] = $this->param('__replyto');
            }
```
image

Hanawa02:
Added the option to set the reply-to mailer option through parameters in the Forms API.

Closes  #1453.

### DIFF
--- a/modules/Forms/Controller/RestApi.php
+++ b/modules/Forms/Controller/RestApi.php
@@ -28,6 +28,9 @@ class RestApi extends \LimeExtra\Controller {
                 $options['subject'] = $this->param('__mailsubject');
             }
 
+            if ($this->param('__replyTo')) {
+                $options['reply_to'] = $this->param('__replyTo');
+            }
             return $this->module('forms')->submit($form, $data, $options);
         }
 


aheinze:
Hi 👋  Thank you for your contribution and sorry for the late reply.

I don't like the idea that you can provide the reply-to via a parameter. For me it is a security concern here. That's why won't merge this PR, sorry.

But I also understand your approach. You can adjust the functionality easily by creating a `config/bootstrap.php` file with the following content:

```php
<?php

$app->on('forms.submit.before', function($form, &$data, $frm, &$options) {

    if (COCKPIT_API_REQUEST && $this->param('__replyTo') && filter_var($this->param('__replyTo'), FILTER_VALIDATE_EMAIL)) {
        $options['reply_to'] = $this->param('__replyTo');
    }
});
```

It is basically binding to the `forms.submit.before` event and modifying the mailer options before sending the email.

Hope this helps!

PR #1457, opened 2021-07-23
Fix (MongoLite): change disallowed character from ( to single quote '

abernh:
Single quotes would allow to break out of the later formed command string allowing any functions to be executed

rel: https://github.com/agentejo/cockpit/commit/b40d6bdedb87265e700ac09007603e72459e7629#commitcomment-53000630


aheinze:
Thanks!

PR #1458, opened 2021-07-23
fix (Auth): prevent username guessing

abernh:
changed error messages for missing user to prevent (brute forced) username guessing


aheinze:
Thanks!

PR #1460, opened 2021-07-27
fix (removeSingleton): remove singleton key also from DB

abernh:
## Bug 
Removing a singleton currently only removes the "singleton spec" file but not its fields data in the DB.

This causes old entries to repopulate a singleton after it was deleted and re-created with the same name and fields.

See also the discussion here https://discourse.getcockpit.com/t/entries-deleted-inside-cockpit-are-retained-in-api/1941/2?u=abernh

## Expected behavior
deleting a singleton should also delete all its data and re-creating it with the same fields should not show the old values of the previous singleton

## Fix
Calling `->storage->removeKey()` on the singleton on-remove solves the issue.


aheinze:
Thanks!

PR #1463, opened 2021-08-12
Improve cookie helper: add samesite parameter

abernh:
# Problem

With the current `Cookie::set()` method the parameter "samesite" can not be set for a cookie.

# Solution

Added additional parameter "samesite" and use the new  `setcookie(name, val, options)` (since PHP7.3) method

# Note

Also renamed misspelled `coockie` helper reference in `Lime\App`:
`'coockie' --> 'cookie'`


aheinze:
Thanks!

wiedaw:
@abernh how can I set secure flag for cookie?

Issue #1002, opened 2019-02-08
Cockpit generates duplicate _id values


rootzjg:
We have integrated Cockpit to our own system. The system creates collection entries based on what we have in our own database. When starting with an empty Cockpit CMS, the initial import creates thousands of collections entries using the API as fast as possible.

This results in duplicate _id values for some (but not all) of the entries that were created at the same second (same timestamp in _created field). 

```
SELECT json_extract(Games5c3340b034ba2.document,'$._id') as id, count(*) as count
FROM Games5c3340b034ba2
GROUP BY json_extract(Games5c3340b034ba2.document,'$._id') 
HAVING count > 1;

5c40b42966623700110001b8|2
5c40b4296662370011000389|2
5c40b42966623700120003b1|2
5c40b42a666237001100011c|2
5c40b42c666237001200012f|2
5c40b42c66623700150000fc|2
5c40b42d666237001400009d|2
5c40b42d66623700140001a5|2
5c40b42d6662370016000245|2
5c40b42e6662370014000237|2
5c40b42e666237001500009f|2
5c40b431666237001300005d|2
5c40b438666237001300028b|2
5c40b438666237001500031d|2
5c40b43966623700120001f2|2
5c40b439666237001300001a|2
5c40b439666237001300024e|2
5c40b43c666237001600039a|2
5c40b44066623700140001c7|2
5c40b4446662370011000377|2
5c40b444666237001200030d|2
```

I checked that these duplicate _id values are found only for entries that have the same creation/modify timestamp. Here's one example (timestamp: 1547744300) where you can see that a couple of entries have the same _id:

```
sqlite> SELECT json_extract(Games5c3340b034ba2.document,'$._id') as id, json_extract(Games5c3340b034ba2.document,'$.gameID') 
   ...> FROM Games5c3340b034ba2
   ...> WHERE json_extract(Games5c3340b034ba2.document,'$._created') = 1547744300
   ...> order by json_extract(Games5c3340b034ba2.document,'$._id');
5c40b42c6662370012000019|434
5c40b42c666237001200001e|7710
5c40b42c666237001200005e|763
5c40b42c666237001200007a|607
5c40b42c666237001200007f|1373
5c40b42c66623700120000ab|193
5c40b42c66623700120000bc|787
5c40b42c66623700120000df|1326
5c40b42c666237001200012f|2209
5c40b42c666237001200012f|4028
5c40b42c6662370012000144|1220
5c40b42c666237001200014c|274
[...]

PR #1465, opened 2021-08-27
fix (Bug 1002:createMongoDbLikeId): microsecond float to big integer …

abernh:
…before binary conversion

microseconds were omitted when using `pack` with format 'N',
causing timestamp to be reduced to "seconds",
causing high chances of equal IDs on import processes

Rel: Issue #1002


aheinze:
Thanks 👍

PR #1473, opened 2021-11-01
Add stale.yml

Raruto:
In order to automatically handle inactive issues: 

- add a label "stale" on issues and pull requests after 30 days of inactivity and comment on them
- close the stale issues after 14 days of inactivity
- if an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart

**Reference:**
- [closing inactive issues](https://docs.github.com/en/actions/managing-issues-and-pull-requests/closing-inactive-issues)
- [actions/stale](https://github.com/actions/stale)

_Have a nice Day,
Raruto_

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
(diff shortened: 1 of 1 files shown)


vanling:
@aheinze, I would definitely do something like this. The amount of issues and PR can scare potential new users away.

PR #1475, opened 2021-11-16
Bugfix entries batch edit localized fields

fredrikkarlsson-pervanovo:
Batch edit is broken in two different ways when working with entries with localized fields:

Since fields (in entries-batchedit.tag) does not contain the localized names, field will be null and the required fields code will have a NPE when saving. Localized fields are added to fields _fields and therefore those should be used to look for localized fields.

When batch editing entries with a selected language, the entries will have empty root fields overwritten by their the selected language field value. The easiest fix is to only allow batch editing while viewing entries in the default language.
The nice fix is to implement patch updating to collection entries and only change the provided fields.
I went for the easy fix this time ;)

### DIFF
--- a/modules/Collections/assets/entries-batchedit.tag
+++ b/modules/Collections/assets/entries-batchedit.tag
@@ -130,8 +130,7 @@
             var required = [], field;
 
             Object.keys(this.checked).forEach(function(name){
-
-                field = $this.fields[name];
+                field = _.find($this._fields, {name: name});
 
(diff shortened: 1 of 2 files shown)


aheinze:
Thanks!

fredrikkarlsson-pervanovo:
Thank you! It would be awesome if you considered some of the other PR's I have made. They have all been battle tested by my organisation since they were created

PR #1476, opened 2021-11-17
Reordering sorted collection is missing a webhook

fredrikkarlsson-pervanovo:
Simple enough, this PR triggers "collection.reorder" and "collections.reorder.{$name}" when a sorted collection is reordered


aheinze:
Thanks!

